PR #27, opened 2020-05-16
Feature/ldap sign in

DerMolly:


src/components/landing/pages/login/auth/via-ldap.tsx
```diff
@@ -1,30 +1,70 @@
...
+                console.log(loginJson)
+                getAndSetUser();
+            }).catch(_reason => {
+                setError(true);
+            }
+        )
+        event.preventDefault();
+    }
+
+    const name = ldapCustomName ? `${ldapCustomName} (LDAP)` : "LDAP";
```
mrdrogdrog, line 27: Why are you adding `(LDAP)` to a custom name?

DerMolly: I didn't change it to before https://github.com/codimd/server/blob/master/public/views/shared/signin-modal.ejs#L60

mrdrogdrog: okay

PR #42, opened 2020-05-17
Add element-separator

mrdrogdrog:
This PR adds a component, which can place an react element between its children.

### DIFF
--- a/src/components/element-separator/element-separator.tsx
+++ b/src/components/element-separator/element-separator.tsx
@@ -1,10 +1,10 @@
 import React, {Fragment} from "react";
 
-export interface HrSeparatorProps {
+export interface ElementSeparatorProps {
     separator: React.ReactElement
 }
 
-export const ElementSeparator: React.FC<HrSeparatorProps> = ({children, separator}) => {
+export const ElementSeparator: React.FC<ElementSeparatorProps> = ({children, separator}) => {
     return (
         <Fragment>
             {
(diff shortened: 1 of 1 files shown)


mrdrogdrog:
An example usage could be
```
<ElementSeparator separator={<hr className="w-100 bg-white"/>}>
    {emailForm}
    {ldapForm}
</ElementSeparator>
```

src/components/element-separator/element-separator.tsx
```diff
@@ -0,0 +1,22 @@
+import React, {Fragment} from "react";
+
+export interface HrSeparatorProps {
```
DerMolly:
```suggestion
export interface ElementSeparatorProps {
```

src/components/element-separator/element-separator.tsx
```diff
@@ -0,0 +1,22 @@
...
+    separator: React.ReactElement
+}
+
+export const ElementSeparator: React.FC<HrSeparatorProps> = ({children, separator}) => {
```
DerMolly:
```suggestion
export const ElementSeparator: React.FC<ElementSeparatorProps> = ({children, separator}) => {
```

src/components/element-separator/element-separator.tsx
mrdrogdrog: Already done.

src/components/element-separator/element-separator.tsx
mrdrogdrog: Already done.

PR #62, opened 2020-05-24
use async and await

mrdrogdrog:
Replace Promise then chains with async / await

### DIFF
--- a/src/components/landing/pages/login/auth/via-ldap.tsx
+++ b/src/components/landing/pages/login/auth/via-ldap.tsx
@@ -31,7 +31,6 @@ const ViaLdap: React.FC = () => {
         event.preventDefault();
     }
 
-
     return (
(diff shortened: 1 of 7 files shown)


src/components/landing/pages/login/auth/via-ldap.tsx
```diff
@@ -9,30 +9,35 @@ import {ApplicationState} from "../../../../../redux";
 const ViaLdap: React.FC = () => {
     const {t} = useTranslation();
     const ldapCustomName = useSelector((state: ApplicationState) => state.backendConfig.customAuthNames.ldap);
+    const name = ldapCustomName ? `${ldapCustomName} (LDAP)` : "LDAP";
```
DerMolly, line 12: let's move that after all use* Operation

PR #75, opened 2020-05-26
Fix title on history page not being translated

ErikMichelson:
The title on the history page was yet untranslated. This PR adds translation support for the title and sets a bottom-margin.


ErikMichelson:
> Please add a translation to at least en.json for the fallback.

The `history` translation-key is already present in almost any locale file. See [en.json](https://github.com/codimd/react-client/blob/dev/public/locales/en.json#L6) for example.

It's the same key that is used for the header nav-buttons (Intro, History).

PR #84, opened 2020-05-29
Feature/history api

DerMolly:


src/api/history.ts
```diff
@@ -0,0 +1,73 @@
...
+export const setHistory = async (entries: HistoryEntry[]): Promise<void> => {
+  const response = await fetch(getBackendUrl() + '/history', {
+    method: 'POST',
+    mode: 'cors',
+    cache: 'no-cache',
+    credentials: 'same-origin',
+    headers: {
+      'Content-Type': 'application/json'
+    },
+    redirect: 'follow',
```
ErikMichelson, line 19: Why should we follow redirects?

mrdrogdrog: Because.. of.. redirecting load balancers? 😅
No srsly. Why shouldn't we?

DerMolly: To be honest I just copied the settings of the other api calls, maybe we should extract default fetch config

DerMolly: extracted

PR #86, opened 2020-05-29
Fix dark mode button

### LINKED ISSUE #83: Add no-focus class for bootstrap button components
The editor, but the css rule might be defined globally.

In the current `feature/editor` branch, the mode-toggle buttons have an outline focus after clicking them.

image

Add a `no-focus` class, that can be applied to these buttons (and maybe other ones).

n/a

**Additional context**
n/a

mrdrogdrog:
This PR adds a new css class, that removes the bootstrap border for focused elements. Fixes #83


mrdrogdrog:
While I tested this PR I used the keyboard to switch through the buttons.
If you do this, the focus becomes very essential.
I don't think we shouldn't disable the focus.

(but I fixed the dark mode button)

PR #106, opened 2020-05-31
Restructure Link Components and fix internal link children error

mrdrogdrog:
This PR fixes a bug in the InternalLinkComponents and restructures the types of the link components.

### DIFF
--- a/src/components/links/external-link.tsx
+++ b/src/components/links/external-link.tsx
@@ -1,18 +1,12 @@
-import React, { Fragment } from 'react'
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
-import { IconProp } from '../../utils/iconProp'
+import React, { Fragment } from 'react'
+import { GeneralLinkProp } from './types'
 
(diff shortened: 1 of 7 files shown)


src/components/links/external-link.tsx
```diff
@@ -1,18 +1,12 @@
...
-  icon?: IconProp;
-  className?: string
-}
+import React, { Fragment } from 'react'
+import { GeneralLinkProp } from './types'
 
-export interface ExternalLinkTextProp {
+export interface ExternalLinkProps extends GeneralLinkProp {
   text: string;
 }
```
DerMolly, line 7: since this and `InternalLinkProps` are basicly the same we maybe should extract that to types.ts too?

DerMolly:
Also we could move the TranslatedLinkProps from translated-external-link.tsx to types.ts

src/components/links/external-link.tsx
mrdrogdrog: true. done.

PR #109, opened 2020-05-31
Restructure redux code

mrdrogdrog:
This PR restructures the redux state code. It also moves some types to other directories.


src/api/frontend-config/types.ts
```diff
@@ -0,0 +1,3 @@
+export interface FrontendConfig {
+  backendUrl: string,
```
ErikMichelson, line 2: There's a tailing comma although no element is following.

mrdrogdrog: removed

src/components/landing/pages/intro/cover-buttons/cover-buttons.tsx
```diff
@@ -1,17 +1,16 @@
...
+import { Link } from 'react-router-dom'
 import { ApplicationState } from '../../../../../redux'
 import './cover-buttons.scss'
 
 export const CoverButtons: React.FC = () => {
   useTranslation()
   const user = useSelector((state: ApplicationState) => state.user)
 
-  if (user.status === LoginStatus.ok) {
+  if (!user) {
```
ErikMichelson, line 13: The logic is inverted here
```suggestion
  if (user) {
```

mrdrogdrog: good point

PR #117, opened 2020-06-01
finished history toolbar buttons

### LINKED ISSUE #45: History Toolbar Part 2 - Reactive Boogaloo
These buttons aren't working:

- [ ] Download history
- [ ] Upload history
- [ ] Refresh history
- [x] Clear history

DerMolly:
fixes #45 
Signed-off-by: Philip Molares <philip.molares@udo.edu>

### DIFF
--- a/src/components/landing/pages/history/history-toolbar/import-history-button.tsx
+++ b/src/components/landing/pages/history/history-toolbar/import-history-button.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react'
+import React, { useRef } from 'react'
 import { Button } from 'react-bootstrap'
 import { useTranslation } from 'react-i18next'
 import { ForkAwesomeIcon } from '../../../../../fork-awesome/fork-awesome-icon'
@@ -11,7 +11,7 @@ export interface ImportHistoryButtonProps {
 
 export const ImportHistoryButton: React.FC<ImportHistoryButtonProps> = ({ onImportHistory }) => {
   const { t } = useTranslation()
-  const [inputRef, setInputRef] = useState<HTMLInputElement | null>()
+  const uploadInput = useRef<HTMLInputElement>(null)
 
   const handleUpload = (event: React.ChangeEvent<HTMLInputElement>) => {
     const { validity, files } = event.target
@@ -34,10 +34,10 @@ export const ImportHistoryButton: React.FC<ImportHistoryButtonProps> = ({ onImpo
   return (
     <div>
       <input type='file' id='upload-history' className="invisible" accept=".json" onChange={handleUpload}
-             ref={element => setInputRef(element)}/>
+        ref={uploadInput}/>
       <Button variant={'light'}
-              title={t('landing.history.toolbar.import')}
-              onClick={() => inputRef?.click()}
+        title={t('landing.history.toolbar.import')}
+        onClick={() => uploadInput.current?.click()}
       >
         <ForkAwesomeIcon icon='upload'/>
       </Button>
--- a/src/components/landing/pages/history/history-toolbar/import-history-button.tsx
+++ b/src/components/landing/pages/history/history-toolbar/import-history-button.tsx
@@ -2,7 +2,8 @@ import React, { useRef, useState } from 'react'
 import { Button, Modal } from 'react-bootstrap'
 import { Trans, useTranslation } from 'react-i18next'
 import { ForkAwesomeIcon } from '../../../../../fork-awesome/fork-awesome-icon'
-import { HistoryEntry } from '../history'
+import { convertOldHistory, OldHistoryEntry } from '../../../../../utils/historyUtils'
+import { HistoryEntry, HistoryJson } from '../history'
 
 export interface ImportHistoryButtonProps {
   onImportHistory: (entries: HistoryEntry[]) => void
@@ -23,7 +24,7 @@ export const ImportHistoryButton: React.FC<ImportHistoryButtonProps> = ({ onImpo
     if (files && files[0] && validity.valid) {
       const file = files[0]
       setFilename(file.name)
-      if (file.type !== 'application/json') {
+      if (file.type !== 'application/json' && file.type !== '') {
         handleShow()
         return
       }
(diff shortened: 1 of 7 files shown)


DerMolly:
needs error handling, when a non-valid file is uploaded in the import function.

src/components/landing/pages/history/history-toolbar/import-history-button.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+export interface ImportHistoryButtonProps {
+  onImportHistory: (entries: HistoryEntry[]) => void
+}
+
+export const ImportHistoryButton: React.FC<ImportHistoryButtonProps> = ({ onImportHistory }) => {
+  const { t } = useTranslation()
+  const uploadInput = useRef<HTMLInputElement>(null)
+  const [show, setShow] = useState(false)
+  const [fileName, setFilename] = useState('')
```
mrdrogdrog, line 15: You're using the filename only in the error modal. How about replacing the `show` state with a null check on `filename` ?

DerMolly: That could be done

src/components/landing/pages/history/history.tsx
```diff
@@ -21,17 +21,36 @@ export const History: React.FC = () => {
...
   useEffect(() => {
-    if (historyEntries === []) {
+    if (!historyEntries || historyEntries === []) {
       return
     }
     setHistoryToLocalStore(historyEntries)
   }, [historyEntries])
 
+  const exportHistory = () => {
+    const data = 'data:text/json;charset=utf-8,' + encodeURIComponent(JSON.stringify(historyEntries))
```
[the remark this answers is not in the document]
DerMolly: can you do that to get 'normal' json files?

DerMolly:
could you provide the crashing file in the 1.6 format, @ErikMichelson ?

mrdrogdrog:
> Tests:
>  - Export history, change one entry and import again ✔
>  - Try to import invalid json file ⇒ show error ✔
>  - Try to import history file in old  (1.6) format ⇒ app crashes without error handling ❌
>  - Try to import arbitrary file ⇒ show error ✔
>  
> In my opinion it would be important to support old files too by converting them into the new format on-the-fly.

People use the export for backups, so I also think, that supporting the old format is important. Maybe we should put versions strings into the export? It might become handy for the next time the format changes.

DerMolly:
Do we want `{ version: '2.0' }` or `{ version: '1.0' }`

mrdrogdrog:
I would use numbers instead of strings, that increment with every change to the format. Less complicated to parse.

mrdrogdrog:
Also nobody is supposed to see this version, so it doesn't have to be human readable.

DerMolly:
Like this:
```
{
  "version":1,
  "entries": …
}
```

src/components/landing/pages/history/history-toolbar/import-history-button.tsx
mrdrogdrog: ok

SISheogorath:
Could we use semver for versions?

JSON representation is:

```JSON
version: {
  major: 1
  minor: 2
  patch: 3
}
```

Of course one could just count, but it has the problem that you suddenly start to setup number ranges if you want to support certain versions later on.

src/components/landing/pages/history/history.tsx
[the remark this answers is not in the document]
ErikMichelson: Ah this is not resolved yet.
It's completely possible. And would look like this:
```javascript
const data = 'data:text/json;charset=utf-8;base64,' + Buffer.from(JSON.stringify(dataObject)).toString('base64')
```

While I'm unsure if it's better to base64-encode the data rather than just replacing special characters, I've seen the base64 approach more often in the wild.

src/utils/historyUtils.ts
```diff
@@ -58,30 +58,35 @@ export function formatHistoryDate (date: Date): string {
...
-    return oldHistory.map((entry: OldHistoryEntry) => {
-      return {
-        id: entry.id,
-        title: entry.text,
-        lastVisited: moment(entry.time).toDate(),
-        tags: entry.tags,
-        pinned: entry.pinned
-      }
-    })
+    const oldHistory = oldHistoryJsonString ? JSON.parse(JSON.parse(oldHistoryJsonString)) as V0HistoryEntry[] : []
```
ErikMichelson, line 88: Wait, we're parsing the JSON twice here? Why?

mrdrogdrog: image

mrdrogdrog: Because codimd stringifies the history array twice.

ErikMichelson: Argh, okay.

src/components/landing/pages/history/history.tsx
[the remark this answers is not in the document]
ErikMichelson: I've done some quick tests with some example history exports and compared the size of the base64 result with the size of the uri-encoded result.

raw JSON: 2797 chars
uri-encoded: 4109 chars
base64: 3732 chars

While this is not such a big difference for this example, the heavy usage of {, } and : characters  in the JSON make the uri-encoded version much larger with larger histories.

mrdrogdrog:
> Could we use semver for versions?
> 
> JSON representation is:
> 
> ```json
> "version": {
>   "major": 1,
>   "minor": 2,
>   "patch": 3
> }
> ```
> 
> Of course one could just count, but it has the problem that you suddenly start to setup number ranges if you want to support certain versions later on.

We discussed this and came to the conclusion, that the version should probably track the api version. Because every breaking change in the history format must also be reflected in the api.

src/utils/historyUtils.ts
DerMolly: yeah, we do that because the old frontend `JSON.stringfy`ed twice

src/components/landing/pages/history/history.tsx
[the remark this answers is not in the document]
DerMolly: I worked that in

PR #128, opened 2020-06-04
show local and remote History

### LINKED ISSUE #46: Load history from server if logged in
If the user is logged is the history component should load the history from the server instead of the local storage.

### LINKED ISSUE #119: Show local and remote history
In version 1.6. you can only see your local history or the remote history of notes.
If you're not logged in, then the history is shown from the local storage. If you're logged in, then you can only see your server history.
But sometimes it's necessary to see your local history while you're logged in, because you edited a note anonymously.

Show the local history, too if you're logged in. (Maybe with a toggle button in the history toolbar?)
If a note is in both histories, then show only one.

You could always log out, find the note and log in again, but that's laborious.

**Additional context**
This issue doesn't deal with things like history migration, when to clear the local history or other writing actions! It's just the possibiliy to see BOTH histories.

DerMolly:
fixes #119 
fixes #46


DerMolly:
Current state of the Card UI:
image

Notice the cloud and laptop icons to indicate the location of a history entry on the left side of the card

Current state of the Table UI:
image

src/components/landing/pages/history/common/pin-button.scss
```diff
@@ -1,12 +1,15 @@
...
-  opacity: 0.2;
-  transition: opacity 0.2s ease-in-out, color 0.2s ease-in-out;
 
-  &:hover {
+  .fa {
+    opacity: 0.2;
+    transition: opacity 0.2s ease-in-out, color 0.2s ease-in-out;
+  }
+
+  &:hover .fa{
```
ErikMichelson, line 8:
```suggestion
  &:hover .fa {
```

src/components/landing/pages/history/common/pin-button.scss
```diff
@@ -1,12 +1,15 @@
...
     opacity: 1;
   }
 
-  &.active {
+  &.pinned .fa{
```
ErikMichelson, line 12: Same as above (missing space)
```suggestion
  &.pinned .fa {
```

src/components/landing/pages/history/common/sync-status.scss
```diff
@@ -0,0 +1,10 @@
+.sync-icon {
+  .fa {
+    opacity: 0.2;
+    transition: opacity 0.2s ease-in-out, color 0.2s ease-in-out;
+  }
+
+  &:hover .fa{
```
ErikMichelson, line 7: Same as above (missing space)
```suggestion
  &:hover .fa {
```

mrdrogdrog:
> Local history gets deleted with the latest changes

solved

src/components/landing/pages/history/history.tsx
```diff
@@ -75,32 +122,45 @@ export const History: React.FC = () => {
...
-    .reduce((a, b) => ([...a, ...b]), [])
-    .filter((value, index, array) => {
-      if (index === 0) {
-        return true
-      }
-      return (value !== array[index - 1])
-    })
-  const entriesToShow = sortAndFilterEntries(historyEntries, viewState)
+  const tags = useMemo<string[]>(() => {
+    return mergeEntryArrays(localHistoryEntries, remoteHistoryEntries).map(entry => entry.tags)
```
DerMolly, line 126: if we have entries to show already shouldn't we maybe use that here?

mrdrogdrog: No, because entriesToShow is already filtered. For the tags we need the unfiltered data.

DerMolly:
If a user is logged in and they try to upload a history json, because of current mock state the upload fails, but in https://github.com/codimd/react-client/blob/051b27e9d7ce9e3057533819e207dc410e8a8245/src/api/history.ts#L11-L20
no error is thrown and therefore 
https://github.com/codimd/react-client/blob/051b27e9d7ce9e3057533819e207dc410e8a8245/src/components/landing/pages/history/history.tsx#L63-L67
won't show that an error has occurred.
I think that maybe a general problem with the api calls right now…

mrdrogdrog:
> If a user is logged in and they try to upload a history json, because of current mock state the upload fails, but in
> 
> https://github.com/codimd/react-client/blob/051b27e9d7ce9e3057533819e207dc410e8a8245/src/api/history.ts#L11-L20
> 
> 
> no error is thrown and therefore
> https://github.com/codimd/react-client/blob/051b27e9d7ce9e3057533819e207dc410e8a8245/src/components/landing/pages/history/history.tsx#L63-L67
> 
> 
> won't show that an error has occurred.
> I think that maybe a general problem with the api calls right now…

should be fixed in https://github.com/codimd/react-client/pull/168

PR #156, opened 2020-06-05
Add function for language code shortening

### LINKED ISSUE #155: Language picker doesn't use default language
**Description**
If you never choose a language manually, then i18next guesses your language with your browser settings. But the language stays on "english"

**To Reproduce**
1. Use a private browser window / delete the i18next entry in the locale storage
2. reload the page

**Expected behavior**
The language picker should select the default language of the browser.

**Screenshots**
image

**Your Setup (please complete the following information):**
 - OS: MacOS Catalina
 - Chrome 81
 - CodiMD version
   - server: mock
   - client: c56430a6a6cdc2981a757d378091ba8e9eb2f9bf

mrdrogdrog:
The browser is sending the combined language Code `language-Area`, but our language file names don't contain the area.
This PR adds a function, that removes the Area from the provided language code.

Fixes #155

### DIFF
--- a/src/components/landing/layout/footer/language-picker.tsx
+++ b/src/components/landing/layout/footer/language-picker.tsx
@@ -13,10 +13,11 @@ const LanguagePicker: React.FC = () => {
 
   const shortenLanguageCode = ():string => {
     const language = i18n.language
-    if (language.substr(0, 2) === 'zh') {
+    const languagePart = language.substr(0, 2)
+    if (languagePart === 'zh') {
       return language
     } else {
-      return language.substr(0, 2)
+      return languagePart
     }
   }
 
--- a/src/components/landing/layout/footer/language-picker.tsx
+++ b/src/components/landing/layout/footer/language-picker.tsx
@@ -11,13 +11,12 @@ const LanguagePicker: React.FC = () => {
     await i18n.changeLanguage(event.currentTarget.value)
   }
 
-  const shortenLanguageCode = ():string => {
-    const language = i18n.language
-    const languagePart = language.substr(0, 2)
-    if (languagePart === 'zh') {
+  const shortenLanguageCode = (language: string):string => {
+    const languageWithoutArea = language.substr(0, 2)
+    if (languageWithoutArea === 'zh') {
       return language
     } else {
-      return languagePart
+      return languageWithoutArea
     }
   }
 
@@ -26,7 +25,7 @@ const LanguagePicker: React.FC = () => {
       as="select"
       size="sm"
       className="mb-2 mx-auto w-auto"
-      value={shortenLanguageCode()}
+      value={shortenLanguageCode(i18n.language)}
       onChange={onChangeLang}
     >
       <option value="en">English</option>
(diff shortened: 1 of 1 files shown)


src/components/landing/layout/footer/language-picker.tsx
```diff
@@ -11,12 +11,21 @@ const LanguagePicker: React.FC = () => {
     await i18n.changeLanguage(event.currentTarget.value)
   }
 
+  const shortenLanguageCode = ():string => {
+    const language = i18n.language
+    if (language.substr(0, 2) === 'zh') {
+      return language
+    } else {
+      return language.substr(0, 2)
```
ccoenen, line 19: why the extra handling of `'zh'`?

src/components/landing/layout/footer/language-picker.tsx
```diff
@@ -11,12 +11,21 @@ const LanguagePicker: React.FC = () => {
...
+    }
+  }
+
```
DerMolly, line 22: this function should have the type `(language: string): string`, it's not good style to get the language from i18n in the function...

src/components/landing/layout/footer/language-picker.tsx
SISheogorath: Because there is zh-tw and zh-cn, but the handling is a bit odd :neutral_face:

src/components/landing/layout/footer/language-picker.tsx
SISheogorath: Could we get an enum for all available languages and return this enum as type? This would prevent invalid return values.

src/components/landing/layout/footer/language-picker.tsx
mrdrogdrog: It's necessary. I need only the language from the 'language-area' code. Except for these two special cases.

src/components/landing/layout/footer/language-picker.tsx
mrdrogdrog: I extracted the i18n reference. 

String as a return type is okay, because `i18n.language` returns a string. I can't verify that with an enum, because enums only exist in typescript. Not in the browser.

src/components/landing/layout/footer/language-picker.tsx
```diff
@@ -11,12 +11,21 @@ const LanguagePicker: React.FC = () => {
     await i18n.changeLanguage(event.currentTarget.value)
   }
 
+  const shortenLanguageCode = (language: string):string => {
```
DerMolly, line 14:
```suggestion
  const shortenLanguageCode = (language: string): string => {
```

mrdrogdrog: done

src/components/landing/layout/footer/language-picker.tsx
ErikMichelson: I would like another variant where these cases are determined automatically too, but this would probably be much more work and at least currently it's not really necessary.

PR #163, opened 2020-06-06
Sort components

mrdrogdrog:
This PR moves all common components that are placed directly in the "components" directory into the "common" directory


DerMolly:
sort-button and pagination should maybe move to the landing/page/history directory? Because they are currently not common and I doubt pagination will become that. Also not sure about sort-button, but maybe it could be used again. Notice I explicitly left icon-button sit in common and not move it to sign-in, because I think we may have use for the icon button again…

mrdrogdrog:
Pagination and SortButton are built to be common. They have no history-specific implementation. They are only used in the history.

PR #165, opened 2020-06-06
Restructure github flow

mrdrogdrog:
This PR restructures the github action flow, so it doesn't use an external action.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -17,9 +17,11 @@ jobs:
         with:
           path: node_modules
           key: node_modules
+      - name: Set up NodeJS
+        uses: actions/setup-node@v1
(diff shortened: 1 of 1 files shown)


SISheogorath:
We should also test node version 10 as we support all NodeJS LTS releases that are still actively maintained.

DerMolly:
Does anybody know how we get the build check to run? Or ignore that and merge anyway? It only appears in the overview on the Conversation Tab but not in the Checks Tab…

mrdrogdrog:
I don't know. Every other PR is green. Maybe the github api where the CI is posting its status is broken again. We had to deal with that at work a few times the last weeks.
@DerMolly please squash merge this PR, because I can't do that.

PR #167, opened 2020-06-06
Changed API path to v2 instead of v2.0

ErikMichelson:
We discussed whether the minor-version is relevant for the API base-path and came to the conclusion, that it's not really needed as breaking API changes need a new major version anyway.
This commit also removes the trailing slash from the URL which is returned by `getBackendUrl`. This is needed as we composed the API routes in our implementation always with a starting slash thus leading to double slashes.

Example of old behaviour:
`getBackendUrl() + '/config'` → example.com/api/v2.0//config

Example of new behaviuor:
`getBackendUrl() + '/config'` → example.com/api/v2/config


ErikMichelson:
@mrdrogdrog The `public/api/v2` is part of this PR, or what should I have forgotten?

mrdrogdrog:
Nevermind. Github has hidden this change.

PR #170, opened 2020-06-07
better support for rtl languages

### LINKED ISSUE #89: Handling RTL
**Description**
For now the UI is not yet supporting right-to-left display mechanism of the UI such as the login page and the top menu.

**Expected behavior**
Depending on the language set in the the user's browser `user-agent` the UI should be displayed the right way such as in arabic, hebrew an farsi languages.

Best regards,

DerMolly:
fixes #89 
 @BoFFire and  @ahangarha should check it out and give feedback if possible…

image

image

### DIFF
--- a/src/components/landing/layout/footer/language-picker.tsx
+++ b/src/components/landing/layout/footer/language-picker.tsx
@@ -2,22 +2,17 @@ import moment from 'moment'
 import React from 'react'
 import { Form } from 'react-bootstrap'
 import { useTranslation } from 'react-i18next'
+import { setHTMLDirection, shortenLanguageCode } from '../../../../initializers/i18n'
 
 const LanguagePicker: React.FC = () => {
   const { i18n } = useTranslation()
 
   const onChangeLang = async (event: React.ChangeEvent<HTMLSelectElement>) => {
-    moment.locale(event.currentTarget.value)
-    await i18n.changeLanguage(event.currentTarget.value)
-  }
-
-  const shortenLanguageCode = (language: string): string => {
-    const languageWithoutArea = language.substr(0, 2)
-    if (languageWithoutArea === 'zh') {
-      return language
-    } else {
-      return languageWithoutArea
-    }
+    const language = event.currentTarget.value
+    console.log('onChangeLang', language)
+    moment.locale(language)
+    await i18n.changeLanguage(language)
+    setHTMLDirection(language)
   }
 
   return (
--- a/src/components/landing/layout/footer/language-picker.tsx
+++ b/src/components/landing/layout/footer/language-picker.tsx
@@ -2,7 +2,7 @@ import moment from 'moment'
 import React from 'react'
 import { Form } from 'react-bootstrap'
 import { useTranslation } from 'react-i18next'
-import { setHTMLDirection, shortenLanguageCode } from '../../../application-loader/initializers/i18n'
+import { shortenLanguageCode } from '../../../application-loader/initializers/i18n'
 
 const LanguagePicker: React.FC = () => {
   const { i18n } = useTranslation()
@@ -11,7 +11,6 @@ const LanguagePicker: React.FC = () => {
     const language = event.currentTarget.value
     moment.locale(language)
     await i18n.changeLanguage(language)
-    setHTMLDirection(language)
   }
 
   return (
--- a/src/components/landing/layout/footer/language-picker.tsx
+++ b/src/components/landing/layout/footer/language-picker.tsx
@@ -2,7 +2,7 @@ import moment from 'moment'
 import React from 'react'
 import { Form } from 'react-bootstrap'
 import { useTranslation } from 'react-i18next'
-import { setHTMLDirection, shortenLanguageCode } from '../../../../initializers/i18n'
+import { setHTMLDirection, shortenLanguageCode } from '../../../application-loader/initializers/i18n'
 
 const LanguagePicker: React.FC = () => {
   const { i18n } = useTranslation()
--- a/src/components/landing/layout/footer/language-picker.tsx
+++ b/src/components/landing/layout/footer/language-picker.tsx
@@ -9,7 +9,6 @@ const LanguagePicker: React.FC = () => {
 
   const onChangeLang = async (event: React.ChangeEvent<HTMLSelectElement>) => {
     const language = event.currentTarget.value
-    console.log('onChangeLang', language)
     moment.locale(language)
     await i18n.changeLanguage(language)
     setHTMLDirection(language)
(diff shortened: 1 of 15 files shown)


ahangarha:
it looks good. but look at this screenshot:

image

I think in the menu, icons should appear before text, so in LTR it should go to left and in RTL it should go to right. That can happen by adding `dir="auto"` to those elements.

DerMolly:
We can't really use `dir=auto` unfortunately, because that only works with the browsers language settings. The App should be able to change via the language picker. Then I just set the whole page to `dir=rtl` if the chosen language is arabic (I think that's our only RTL language right now)...
I could make it that those entries are always ltr, but maybe we should just add an arabic translation for theses texts...

src/initializers/i18n.ts
```diff
@@ -49,4 +49,27 @@ export const setUpI18n = async (): Promise<void> => {
     })
 
   moment.locale(i18n.language)
+  setHTMLDirection(shortenLanguageCode(i18n.language))
+}
+
+export const setHTMLDirection = (language: string): void => {
+  console.log(language)
```
ErikMichelson, line 56: Why do we output/log this?

src/components/landing/layout/footer/language-picker.tsx
```diff
@@ -2,22 +2,17 @@ import moment from 'moment'
...
-
-  const shortenLanguageCode = (language: string): string => {
-    const languageWithoutArea = language.substr(0, 2)
-    if (languageWithoutArea === 'zh') {
-      return language
-    } else {
-      return languageWithoutArea
-    }
+    const language = event.currentTarget.value
+    console.log('onChangeLang', language)
```
ErikMichelson, line 12: Same question as below: Why do we output/log this? This isn't production relevant and bloating the code unnecessary.

src/initializers/i18n.ts
DerMolly: that's 'debug' code

src/components/landing/layout/footer/language-picker.tsx
DerMolly: that's 'debug' code

src/initializers/i18n.ts
ErikMichelson: Yes, and debug-code should not be committed, should it?

mrdrogdrog: dont use console.log. Use console.debug please.

ahangarha:
@DerMolly 

See. As a person whose primary languge is Persian which is RTL, there are many cases in which I have LTR texts. Beside this, There are some cases in which, here are English texts in the app which are not translated or not supposed to be translated.

I don't know what is the issue for not adding `dir="auto"` but If you want, I can send a PR. Just I don't know from where the dropdown menu gets `text-align: left;`. I can manage it by adding `text-align: start;` for the relevant elements in styles.

Let me know what to do.

DerMolly:
I think I misunderstood what exactly `dir='auto'` does. I added it to a toplevel div and some elements which have intentionally english text or are just not yet translated. Best case scenario we would add  `dir='auto'` to each element via css, but that's unfortunately not possible. At least that's my understanding of https://github.com/w3c/csswg-drafts/issues/2091. 

We should add `dir='auto'` to more elements, to ensure a better translation progress for our next RTL language

mrdrogdrog:
You don't have to add it to every element. The "dir" attribute will be inherited!

DerMolly:
Since `dir='auto'` performes a language direction detection and we could have untranslated texts, we need to add `dir='auto'` not to every element (that would have just been the easiest way if it could be done via CSS), but to each element where a direction change may occur. For arabic that is currently done and works fine (I think), but new RTL languages in the future, which are not 100% translated yet, would need to have extra `dir='auto'` thrown into the mix…

mrdrogdrog:
No, you don't have to put auto to a lower element, if a dir=auto has it already. All we have to do is to set dir=auto to the root div

mrdrogdrog:
image

mrdrogdrog:
Or is there any problem with this representation?

DerMolly:
There is in the menu as @ahangarha pointed out the first to paragraphs should have the icon on the left, as it currently is in this branch and because the text direction changes from rtl to ltr in this case you either need to add `dir='ltr'` (as done for the CodiMD and logo on the intro page) or (because we don't know if the text is truly translated) add an extra `dir='auto'` to those entries to possibly change the text direction.

mrdrogdrog:
So you want mixed directions?

DerMolly:
I think it needs to be possible

mrdrogdrog:
Well... That could be difficult, but not impossible. 
Not every component we use (especially from react-bootstrap) has a direction prop. But we can wrap the content of these elements in e.g. spans and add a direction to them.

DerMolly:
The question is do we want to do that now or do we want to work the scout approach?

mrdrogdrog:
IMHO : I would wait. Currently we are altering the UI very often. It's easier to do such extra tasks later. But I would still include it into the 2.0.

mrdrogdrog:
First I would work out the complete 2.0 UI and then optimize it for RTL.

DerMolly:
Then this should be merged and an extra issue should be opened to fix this when the code is final or at least close to that. In the mean time new code should include `dir='auto'`. I'll merge that later, in case there are other voices to heard here and open the issue.

mrdrogdrog:
Can't we just leave this issue open and add `dir=auto` to the HTML tag for now?

mrdrogdrog:
Please rebase the pr first

ahangarha:
Guys,

I explained above that unless you are sure about whole website contains only LTR or RTL text, it is mistake to think applying direction in top level (root) is going to give desirable result.

Virtually, all the text containing elements should have `dir="auto"` so based on its content, it would show the text in proper direction. This is important to have proper bidi support.

If CSS has `direction:auto;` it would be so easy to implement. But for now, using `dir="auto"` is the only way. The other workaround is to do something similar to what my firefox addon does: run JS on page and add `dir="auto"` and `text-align: start;` to all p, a, hx, ul, ol, li, pre.

mrdrogdrog:
That's why I propose, that we first finish the UI and then, as one of the last tasks for 2.0, add optimizations for e.g. rtl.

ahangarha:
Then keep `dir="auto"` in top level. I try to gradually come and contribute to bidi support.

SISheogorath:
Since react is a component-based structure, shouldn't each component become RTL-aware on its own? In this case PRs could happen for each component separately and therefore being less likely to cause problems with changes.

PR #190, opened 2020-06-11
Adds an info banner to the app

### LINKED ISSUE #189: Add info banner
The top of the page

The admin should have the possibility to add a banner at the top of the client to inform the users about stuff.

DerMolly:
### Component/Part
Github pull request template

### Description
This adds a banner to the top of the landing-layout and the editor. When `banner.text` in the config call is set and the particular banner wasn't dismissed by the user (more precisely their browser) already, the banner is shown. It links to `/n/banner`. If it is dismissed the timestamp is saved in the localStorage and the exact same banner should never show again.

### Steps

- [x] added implementation
- [x] added / updated tests
- [x] extended changelog

### Related Issue(s)
#189


public/api/v2/config
```diff
@@ -13,6 +13,7 @@
     "email": true,
     "openid": true
   },
+  "bannerText": "Dies ist ein Testbanner",
```
SISheogorath, line 16: This should match the banner note and therefore be English.

public/api/v2/notes/banner
```diff
@@ -0,0 +1,14 @@
...
+    "lastChange": {
+        "userId": "snskxnksnxksnxksnx",
+        "username": "testy",
+        "timestamp": 123456789
+    },
+    "viewcount": 0,
+    "createtime": "2020-05-22T20:46:08.962Z",
+    "content": "This is the test banner text",
+    "authorship": [],
+    "preVersionTwoNote": true
```
SISheogorath, line 13: This should be false.

public/api/v2/config
DerMolly: As this is just a mock call not really necessary to change, but technically correct

public/api/v2/notes/banner
DerMolly: As this is just a mock call not really necessary to change, but technically correct

public/api/v2/config
SISheogorath: While I agree that it's not necessary it would be great when mock data would be consistent.

public/api/v2/notes/banner
SISheogorath: If this stays false, I can definitely live with it, but since we should touch the code again anyway, why not fix it all along?

mrdrogdrog: I don't get what the problem is.

src/components/application-loader/initializers/configLoader.ts
```diff
@@ -17,5 +19,16 @@ export const loadAllConfig: (baseUrl: string) => Promise<void> = async (baseUrl)
   }
   setBackendConfig(backendConfig)
 
+  const currentText = backendConfig.bannerText
+  if (currentText !== '') {
+    const bannerNote = await getNote('banner')
+    const lastAcknowledgedTimestamp = window.localStorage.getItem('bannerTimeStamp') || ''
+    setBanner({
+      text: currentText,
+      timestamp: bannerNote.createtime,
```
mrdrogdrog, line 28: I don't think, that we should use `createTime`. If you want to change the banner, then you'll have to delete the note and recreate it. Can we use another attribute? `updateTime`? content? And If we already provide the preview text via backend config.. why can't we provide the timestamp. too?

DerMolly: Re: Providing the timestamp
We can totally do that, too. But we need the banner redux state anyway to keep track of a dismissed banner in one session. So I tried to minimize the stuff in the backend config. We could simply use the first header of the banner.md file as text and throw it out of backend config entirely, but that would lead to an extra request on every full reload. As opposed to only making that request, when the banner text is set. (That basically makes the text a flag and we could replace it with a simple boolean and get the timestamp and banner text via the request if you prefer.)

Re: updateTime
That is something I was thinking too, currently the note for whatever reason has no updateTime field, but I think we should add that.

public/api/v2/notes/banner
DerMolly: @mrdrogdrog As this technically is not a preVersionTwoNote the field should be false, but I just copied the `old` note which is (like the name indicates) supposed to be an old pre 2.0 note.
I'll change that quickly.

mrdrogdrog: Oh okay. I get it. The problem isn't the attribute itself, but the value of the attribute in the banner note.

src/components/application-loader/initializers/configLoader.ts
mrdrogdrog: I don't want the remove the redux state. I want to reduce the number of Ajax calls that are needed to initialize the app

ErikMichelson: It's always a great idea to reduce the amount of startup HTTP requests, therefore I agree that we should include the bannerTimestamp into the config API call.

But in general we should be careful not to totally mix things up by putting everything into an "all-in-one" API-call just to reduce the amount of requests. But as I said above, in this case it's really useful to do so.

PR #192, opened 2020-06-11
Updated schedule of renovate

ErikMichelson:
Even if renovate only creates 1-2 pull requests per day, they start to become annoying. This pull requests changes the time slot of renovate actions to Saturdays.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -33,6 +33,7 @@
       "groupName": "JS test packages"
     }
   ],
+  "prHourlyLimit": 0,
   "schedule": [
     "on Saturday"
   ]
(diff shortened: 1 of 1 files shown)


SISheogorath:
@ErikMichelson you might want to increase the number of PRs allowed per hour, iirc they are limited to 2 by default.

ErikMichelson:
[The documentation](https://docs.renovatebot.com/configuration-options/#prhourlylimit) says the default value is 0, which means unlimited. Or is this the wrong config option?

SISheogorath:
Yes, but we use `config:base` preset, that adds this limit: 

https://docs.renovatebot.com/presets-config/#configbase -> https://docs.renovatebot.com/presets-default/#prhourlylimit2

PR #199, opened 2020-06-11
added redirector component

DerMolly:
it will redirect every request to /$something that is not handled otherwise (/intro, /login and such) to /n/$something

Signed-off-by: Philip Molares <philip.molares@udo.edu>

### DIFF
--- a/public/api/v2/notes/old
+++ b/public/api/v2/notes/old
@@ -1,0 +1,14 @@
+{
+    id: "ABC123",
+    alias: "old",
+    lastChange: {
+        userId: "snskxnksnxksnxksnx",
+        username: "testy",
+        timestamp: 123456789
+    },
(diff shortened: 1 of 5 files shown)


ErikMichelson:
I don't think that redirecting everything to /n/ is a great idea. If I remember correctly the idea in one of the last community calls was to only redirect pre-2.0-notes to the new namespace.
For all other requests that don't match an existent route, I would throw a 404.

DerMolly:
How would we identify only pre-2.0-notes and nothing else?

ErikMichelson:
The idea was to add another column to the notes table that indicates whether a note is created before the 2.0 update of after that.

DerMolly:
I did not remember that, but that's a good idea I will add the according code.

PR #235, opened 2020-06-20
Add codimd to jetbrains dictionary

mrdrogdrog:
### Component/Part
JetBrains-IDE-Settings

### Description
This PR adds the words "codimd" and "CodiMD" to the jetbrains dictionary

### DIFF
--- a/.idea/dictionaries/tilman.xml
+++ b/.idea/dictionaries/tilman.xml
@@ -1,0 +1,8 @@
+<component name="ProjectDictionaryState">
+  <dictionary name="tilman">
+    <words>
+      <w>CodiMD</w>
+      <w>codimd</w>
+    </words>
+  </dictionary>
+</component>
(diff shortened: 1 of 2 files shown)


SuperSandro2000:
If you are all using jetbrains this is fine but if we are split between vim, vscode, jetbrains and 5 more than we shouldn't add this for every editor and just use the local dictionary instead.

mrdrogdrog:
We (@DerMolly , @ErikMichelson  and myself) are using Jetbrains Webstorm for this Project. So it would be handy for us.

.idea/dictionaries/tilman.xml
```diff
@@ -0,0 +1,8 @@
+<component name="ProjectDictionaryState">
+  <dictionary name="tilman">
```
DerMolly, line 2: this name seems not correct

Issue #244, opened 2020-06-20
Re-add yaml metadata parsing


ErikMichelson:
**Describe the solution you'd like**
YAML headers aren't rendered in markdown but extracted and parsed so that the renderer can use the options of it.

**Additional context**
There's an library that seems to be useful for this: https://www.npmjs.com/package/markdown-it-front-matter

ErikMichelson:
This was done in #310

PR #248, opened 2020-06-21
add YAML Metadata

DerMolly:
### Component/Part
markdown renderer

### Description
This PR adds the ability to parse YAML Metadata at the beginning of the markdown file

### Steps

- [x] added implementation
- [x] added / updated tests
- [x] extended changelog

### Related Issue(s)
#244


mrdrogdrog:
I think we should use deep-equal instead of the self made function

DerMolly:
further reading about deep-equal vs. json.stringfy https://www.mattzeunert.com/2016/01/28/javascript-deep-equal.html

DerMolly:
Maybe this can help with the open graph types https://github.com/beau6183/ts-open-graph-scraper/blob/master/src/types.ts

ErikMichelson:
I thought about letting them as an `Map<string,string>` since we don't access these og-attributes directly.

ErikMichelson:
As the code of this PR was heavily changed in #310, it doesn't make sense to continue here. The changes will be included in #310 instead.

PR #268, opened 2020-06-24
Add linkify

mrdrogdrog:
Signed-off-by: Tilman Vatteroth <tilman.vatteroth@tu-dortmund.de>

### Component/Part
The markdown renderer

### Description
This PR reactivates linkify, but executes it at the end of the plugin pipeline

### Steps

- [x] added implementation
- [x] added / updated tests
- [x] extended changelog

### Related Issue(s)
Closes #236

### DIFF
--- a/src/components/editor/markdown-renderer/markdown-it-plugins/linkify-extra.ts
+++ b/src/components/editor/markdown-renderer/markdown-it-plugins/linkify-extra.ts
@@ -1,0 +1,13 @@
+import MarkdownIt from 'markdown-it/lib'
+import linkify from 'markdown-it/lib/rules_core/linkify'
+
+export const linkifyExtra: MarkdownIt.PluginSimple = (md) => {
+  md.core.ruler.push('linkify', state => {
+    try {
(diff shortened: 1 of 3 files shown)


src/components/editor/markdown-renderer/markdown-it-plugins/linkify-extra.ts
```diff
@@ -0,0 +1,13 @@
+import MarkdownIt from 'markdown-it/lib'
+import linkify from 'markdown-it/lib/rules_core/linkify'
```
DerMolly, line 2: we could simply use [llinkify-it](https://www.npmjs.com/package/linkify-it) and then use [@types/linkify-it](https://www.npmjs.com/package/@types/linkify-it) instead of our own type. Not really sure if that's necessary, but it's possible

mrdrogdrog: linkify-it is not the markdown plugin. And I don't want to reinvent the markdown plugin.

PR #278, opened 2020-06-25
Fixed bug that two p-tags were encapsulated inside each other

ErikMichelson:
### Component/Part
Markdown renderer

### Description
This PR fixes a bug that two p-tags were encapsulated inside each other with the same key.
This bug was introduced because the wider-possible-replacer replaced each element that was subject to be enlarged with a brand new p-surrounding. That surrounding got the key of the old (now inner) element and therefore casts a duplicate key exception.
This fix solves that problem by setting the 'wider-possible' class attribute directly on the selected tag instead of creating a surrounding.

### Steps

- [x] added implementation
- [x] added / updated tests
- [x] extended changelog

### Related Issue(s)
none

### DIFF
--- a/src/components/editor/markdown-renderer/replace-components/possible-wider/possible-wider-replacer.tsx
+++ b/src/components/editor/markdown-renderer/replace-components/possible-wider/possible-wider-replacer.tsx
@@ -17,10 +17,7 @@ export class PossibleWiderReplacer implements ComponentReplacer {
     if (!(childIsImage || childIsYoutube || childIsVimeo || childIsPDF)) {
       return
     }
-    if (!node.attribs) {
-      return
-    }
-    node.attribs.class = `wider-possible ${node.attribs.class || ''}`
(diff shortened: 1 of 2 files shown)


src/components/editor/markdown-renderer/replace-components/possible-wider/possible-wider-replacer.tsx
```diff
@@ -17,8 +17,7 @@ export class PossibleWiderReplacer implements ComponentReplacer {
     if (!(childIsImage || childIsYoutube || childIsVimeo || childIsPDF)) {
       return
     }
-    return (<p className='wider-possible'>
-      {subNodeConverter(node, index)}
-    </p>)
+    node.attribs = Object.assign(node.attribs || {}, { class: `wider-possible ${node.attribs?.class || ''}` })
```
DerMolly, line 20: maybe a comment what that does would be a good idea

PR #297, opened 2020-06-27
Replace links to /features with /n/features

ErikMichelson:
### Component/Part
Links in the landing page

### Description
It was discussed and agreed that all notes should reside under the /n/ namespace in the future. Even if we want to redirect all pre-2.0-notes from the root namespace to the /n/ namespace, it might be better to set the internal links for the features page properly instead on relying on the redirect.
This PR therefore replaces each occurence of `/features` with `/n/features`.

### Steps

- [x] added implementation
- [x] added / updated tests
- [ ] extended changelog

### Related Issue(s)
none

### DIFF
--- a/public/locales/en.json
+++ b/public/locales/en.json
@@ -133,7 +133,7 @@
         }
     },
     "editor": {
-        "placeholder": "← Start by entering a title here\n===\nVisit /features if you don't know what to do.\nHappy hacking :)",
+        "placeholder": "← Start by entering a title here\n===\nVisit /n/features if you don't know what to do.\nHappy hacking :)",
         "help": {
             "contacts": {
                 "title": "Contacts",
--- a/public/locales/en.json
+++ b/public/locales/en.json
@@ -133,7 +133,7 @@
         }
     },
     "editor": {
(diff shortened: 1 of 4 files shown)


public/locales/en.json
```diff
@@ -133,7 +133,7 @@
         }
     },
     "editor": {
-        "placeholder": "← Start by entering a title here\n===\nVisit /features if you don't know what to do.\nHappy hacking :)",
+        "placeholder": "← Start by entering a title here\n===\nVisit /n/features if you don't know what to do.\nHappy hacking :)",
```
SISheogorath, line 136: Maybe talk about "the features page" instead, the `/n/` reads a bit odd in this sentence :)

mrdrogdrog, line 136: is that necessary since /features redirects to /n/features?

ErikMichelson: Is the change mentioned by @SISheogorath okay for you?

You're right with the redirect, you could apply that comment to the whole PR. I just thought why should our links in the application have the overhead of a redirect, when they could point to the right target directly. So I did a quick regex replace.

Personally I like "the features page" even more than just "/features", but if you don't like that, I can change that back.

DerMolly: I think this PR is a good idea and I for one like 'the features page' also more than `/n/features`

PR #299, opened 2020-06-27
Add note deletion and removal from history modals

ErikMichelson:
### Component/Part
History -> Note EntryMenu

### Description
This PR adds the confirmation modals when clicking "delete this note" or "remove this note from history" in the context menu of a note. This PR also simplifies some code constructs in the same files.

### Steps

- [x] added implementation
- [x] added / updated tests
- [x] extended changelog

### Related Issue(s)
none


public/locales/en.json
```diff
@@ -70,6 +70,12 @@
                     "textWithFile": "While trying to import history from '{{fileName}}' an error occurred.",
                     "textWithoutFile": "You did not provide any files to upload the history from.",
                     "tooNewVersion": "The file '{{fileName}}' comes from a newer client and can't be imported."
+                },
+                "removeNote": {
+                  "title": "Remove note from history",
+                  "question": "Do you really want to remove this note from your history?",
```
DerMolly, line 76: Do we maybe include the name of the note here?

public/locales/en.json
```diff
@@ -252,7 +258,8 @@
             "deleteNote": {
                 "title": "Delete note",
                 "question": "Do you really want to delete this note?",
```
DerMolly, line 260: same here

src/components/landing/pages/history/common/entry-menu/entry-menu.tsx
```diff
@@ -40,17 +44,11 @@ const EntryMenu: React.FC<EntryMenuProps> = ({ id, location, isDark, onRemove, o
...
-        </Dropdown.Item>
+        <RemoveNoteEntryItem onRemove={onRemove} />
 
         <Dropdown.Divider/>
 
-        <Dropdown.Item onClick={onDelete}>
-          <ForkAwesomeIcon icon="trash" fixedWidth={true} className="mx-2"/>
-          <Trans i18nKey="landing.history.menu.deleteNote"/>
-        </Dropdown.Item>
+        <DeleteNoteItem onDelete={onDelete} />
```
DerMolly, line 51: Aren't those the basically the same, but differ only in the i18nKeys, the callback and the icon?

mrdrogdrog:
Please rebase this PR.

src/components/landing/pages/history/common/entry-menu/entry-menu.tsx
ErikMichelson: You're right with that. I thought about even more abstraction too, but didn't do that as the exact `DeleteNoteItem` component can be used in the dropdown menu of the editor. Furthermore, I'm unsure if we want to have a component with that many attributes (because each i18nKey would need to be an attribute with higher abstraction).

DerMolly: Since you wrote it, you're the expert on that topic and I trust your instincts there. It was just something I noticed while reviewing this

ErikMichelson: What about putting them together into one component that has an option `variant = 'deleteNote' | 'removeHistoryItem'` ? The component would then use the right icon and i18nKeys by switching them internally.

public/locales/en.json
ErikMichelson: Done

public/locales/en.json
ErikMichelson: Done, even if I'm a bit unsure about that, because the same modal is used in the editor's dropdown menu and from the editor the saying 'this note' is more clearly.

src/components/landing/pages/history/common/entry-menu/entry-menu.tsx
ErikMichelson: Done that as described in the answer above.

src/components/landing/pages/history/common/entry-menu/delete-remove-note-item.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { DeletionModal } from '../../../../../common/modals/deletion-modal'
+import { ShowIf } from '../../../../../common/show-if/show-if'
+
+export interface DeleteRemoveNoteItemProps {
+  onConfirm: () => void
+  noteTitle: string
+  variant: 'removeHistoryItem' | 'deleteNote'
+}
+
+export const DeleteRemoveNoteItem: React.FC<DeleteRemoveNoteItemProps> = ({ variant, onConfirm, noteTitle }) => {
```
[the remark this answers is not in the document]
DerMolly: you maybe want to look at the last thread in my earlier review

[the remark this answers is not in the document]
mrdrogdrog: Sry but I think it's better to split components that are meant for different semantic tasks.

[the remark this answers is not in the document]
DerMolly: how about this then, have two components that only use on common component with all the necessary props. That way there are two components, but since they use exactly the same logic the base is the same and we minimize repeating code.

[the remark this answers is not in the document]
mrdrogdrog: The react way to do this is composition. It's totally legit to create an abstract component with many props and then a specific component that just wraps this abstract component with values.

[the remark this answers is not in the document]
DerMolly: Isn't that what I described?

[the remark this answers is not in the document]
mrdrogdrog: Yes. Github hasn't shown your comment when I wrote mine.

ErikMichelson:
Looks like this with the newest commit:
image

PR #311, opened 2020-07-08
Add codemirror keybindings and addons

DerMolly:
### Component/Part
markdown editor

### Description
This PR adds codemirror addons:
- fullScreen
- ~continueComment~ (this does not do anything in markdown documents)
- autorefresh
added a default for:
- extraKeys

### Steps

- [ ] added implementation
- [ ] added / updated tests
- [ ] extended changelog


DerMolly:
`continueComments` seems to be unnecessary

DerMolly:
there seems to be a problem in key-map.ts currently, not sure how to fix that currently.

ErikMichelson:
Is the fullscreen addon still not correctly working because of the way the codemirror editor is embedded? If so, I suggest to just drop that. Any user may switch to 'only edit'-view and may toggle the browser's fullscreen and have almost the same editor space.

DerMolly:
I'm not sure if that's the way or if we should try to preserve the fullscreen mode. But yeah if we can't get it to work, removing it is the way…

DerMolly:
I did just play around with that and the problem seems to be that the `CodeMirror` css class defines `position: relative;` and the fullscreen addon defines `CodeMirror-fullscreen` with `position: fixed;`. Both are applied to the same `<div>`. The question is is that a problem of `codemirror` (I somewhat doubt that, but could be) or `react-codemirror2`? Since the `CodeMirror` and `CodeMirror-fullscreen` classes both come from the codemirror npm package I'm a bit unsure how to fix this…

src/components/editor/editor-window/key-map.ts
```diff
@@ -0,0 +1,111 @@
...
+  // },
+  // 'Shift-Ctrl-=': cm => {
+  //   utils.wrapTextWith(this.editor, cm, '+')
+  // },
+  // 'Ctrl-=': cm => {
+  //   utils.wrapTextWith(this.editor, cm, '=')
+  // },
+  // 'Shift-Ctrl-Backspace': cm => {
+  //   utils.wrapTextWith(this.editor, cm, 'Backspace')
+  // }
```
mrdrogdrog, line 110: Is this code unneeded or not ready yet?

DerMolly: not ready yet

DerMolly: I'm currently debating to use the function `wrapTextWith` not only here but also for the toolbar as this seems to using the codemirror replace function directly which is hopefully better than the custom function over the whole content we're currently using. I will investigate more later and see if I can make it work…

src/components/editor/editor-window/key-map.ts
```diff
@@ -0,0 +1,109 @@
...
+    wrapTextWith(cm, '^')
+  },
+  'Ctrl-+': cm => {
+    wrapTextWith(cm, '+')
+  },
+  'Shift-Ctrl-=': cm => {
+    wrapTextWith(cm, '+')
+  },
+  'Ctrl-=': cm => {
+    wrapTextWith(cm, '=')
```
ErikMichelson, line 107: Why do you wrap the text with a single equality sign? This does nothing to the rendering. If you want a keybinding for hightlighting text, use double equality signs.

src/components/editor/editor-window/key-map.ts
```diff
@@ -0,0 +1,109 @@
...
+      cm.setOption('fullScreen', false)
+    } else {
+      return CodeMirror.Pass
+    }
+  },
+  'Cmd-S': function () {
+    return CodeMirror.Pass
+  },
+  'Ctrl-S': function () {
+    return CodeMirror.Pass
```
ErikMichelson, line 21: This just lets the keybinding pass forward to the browser and does not prevent the saving dialog. Same applies to the `Cmd-S` shortcut.

src/components/editor/editor-window/key-map.ts
```diff
@@ -0,0 +1,109 @@
+import CodeMirror, { KeyMap } from 'codemirror'
+import { wrapTextWith } from './tool-bar/utils'
+
+// const isMac = CodeMirror.keyMap.default === CodeMirror.keyMap.macDefault
```
ErikMichelson, line 4: Is this needed or can this safely be deleted?

src/components/editor/editor-window/key-map.ts
DerMolly: I just copied it from the old code

https://github.com/codimd/server/blob/master/public/js/lib/editor/index.js#L119-L121

, but I'll change it

src/components/editor/editor-window/key-map.ts
DerMolly: It was needed in the old code, but it's so stupid that I removed it, but I think we may need to readd this. Maybe we should rework the shortcuts altogether and remove the old ones for now?

mrdrogdrog: Please let's try to build a clean solution

src/components/editor/editor-window/editor-window.tsx
```diff
@@ -25,42 +28,15 @@ export interface EditorWindowProps {
...
     <div className={'d-flex flex-column h-100'}>
       <ToolBar
-        content={content}
-        onContentChange={onContentChange}
-        positions={positions}
+        editor={editor}
       />
       <ControlledCodeMirror
-        className="overflow-hidden w-100 flex-fill"
+        className="h-100 w-100 flex-fill"
```
mrdrogdrog, line 39: Why did you remove the `overflow-hidden` ? It's important for the scrollbar

src/components/editor/editor-window/key-map.ts
mrdrogdrog: What was its purpose?

src/components/editor/editor-window/tool-bar/tool-bar.tsx
```diff
@@ -1,82 +1,86 @@
...
-      </Button>
-      <Button variant='light' onClick={addLine} title={t('editor.editorToolbar.line')}>
-        <ForkAwesomeIcon icon="minus"/>
-      </Button>
-      <Button variant='light' onClick={addComment} title={t('editor.editorToolbar.comment')}>
-        <ForkAwesomeIcon icon="comment"/>
-      </Button>
-    </ButtonToolbar>
-  )
+  if (editor) {
```
mrdrogdrog, line 34: invert pls.

src/components/editor/editor-window/key-map.ts
DerMolly: fixed

src/components/editor/editor-window/key-map.ts
DerMolly: fixed

src/components/editor/editor-window/key-map.ts
DerMolly: it was used on some shortcuts to differentiate between macs and non-macs, but the latest changes should have made that obsolete

src/components/editor/editor-window/tool-bar/tool-bar.tsx
DerMolly: done

src/components/editor/editor-window/editor-window.tsx
DerMolly: done

src/components/editor/editor-window/key-map.ts
DerMolly: As this is long gone code, could you close this @mrdrogdrog?

src/components/editor/editor-window/key-map.ts
```diff
@@ -0,0 +1,93 @@
+import CodeMirror, { Editor, KeyMap, Pass } from 'codemirror'
+import {
+  makeSelectionBold,
+  makeSelectionItalic,
+  markSelection,
+  strikeThroughSelection,
+  underlineSelection
+} from './tool-bar/utils'
+
+const isMac = navigator.platform.includes('Mac')
```
mrdrogdrog, line 10:
```suggestion
const isMac = navigator.platform.toLowerCase().includes('mac')
```
I think this wouldn't hurt. We never know if every browser will use "Mac".

src/components/editor/editor-window/key-map.ts
```diff
@@ -0,0 +1,93 @@
...
+
+const f10 = (editor: Editor): void | typeof Pass => editor.setOption('fullScreen', !editor.getOption('fullScreen'))
+const esc = (editor: Editor): void | typeof Pass => {
+  if (editor.getOption('fullScreen') && !isVim(editor.getOption('keyMap'))) {
+    editor.setOption('fullScreen', false)
+  } else {
+    return CodeMirror.Pass
+  }
+}
+const supressSave = (): undefined => undefined
```
mrdrogdrog, line 22: sup<b>p</b>ress

src/components/editor/editor-window/key-map.ts
DerMolly: Well according to the Living HTML Standard the value should currently always be 'MacIntel'

src/components/editor/editor-window/key-map.ts
DerMolly: done

src/components/editor/editor-window/key-map.ts
```diff
@@ -0,0 +1,93 @@
...
+  } else {
+    if (editor.getOption('indentWithTabs')) {
+      editor.execCommand('defaultTab')
+    } else {
+      editor.replaceSelection(spaces)
+    }
+  }
+}
+
+export const defaultKeyMap: KeyMap = isMac ? {
```
ErikMichelson, line 65:
```suggestion
export const defaultKeyMap: KeyMap = !isMac ? {
```

The ternary condition needs to be inverted.

PR #315, opened 2020-07-10
Add support for image-proxies by (optional) piping of image urls through the backend

ErikMichelson:
### Component/Part
Editor -> markdown-renderer -> image-frame

### Description
This PR adds support for image proxies by passing image urls through the backend if the corresponding option is enabled in the config.

### Steps

- [x] added implementation
- [x] added / updated tests
- [x] extended changelog

### Related Issue(s)
codimd/server#95


SuperSandro2000:
Some thoughts:
- Does this limit max size? Can I try to proxy a 100GB test file and bring down the entire server?
- What happens with content types? What if I load something that is not an image? Can I inject stuff like this?
- What if the URL I request does an infinite redirect?
- Is the proxy open? Can anyone use it for anything?

Maybe we should utilize some existing software for this and avoid most security questions.

SISheogorath:
You misunderstood this feature, @supersandro2000. This implementation only requests and URL from the back end that provides the picture proxied.

On the server side we already discussed that we'll use camp for this task https://github.com/codimd/server/issues/95

And all those questions are addressed by the camp configuration.

SuperSandro2000:
Oh, yeah. nevermind.

CHANGELOG.md
```diff
@@ -32,6 +32,7 @@
 - Images, videos, and other non-text content is now wider in View Mode
 - Notes may now be deleted directly from the history page
 - CodiMD instances can now be branded either with a '@ <custom string>' or '@ <custom logo>' after the CodiMD logo and text
+- The CodiMD frontend supports now optional image proxies defined in the backend.
```
mrdrogdrog, line 35:
```suggestion
- The CodiMD frontend now supports optional image proxies defined in the backend.
```

src/api/imageProxy.ts
```diff
@@ -0,0 +1,15 @@
+import { ImageProxyResponse } from '../components/editor/markdown-renderer/replace-components/image/types'
+import { expectResponseCode, getBackendUrl } from '../utils/apiUtils'
+import { defaultFetchConfig } from './default'
+
+export const getProxiedUrl = async (imageUrl: string): Promise<ImageProxyResponse> => {
```
mrdrogdrog, line 5: Why are we using a call that delivers an image URL? Why can't we send the image url to the backend and receive an image?

src/components/editor/markdown-renderer/replace-components/image/image-frame.tsx
```diff
@@ -1,7 +1,25 @@
-import React from 'react'
+import React, { useEffect, useState } from 'react'
+import { useSelector } from 'react-redux'
+import { getProxiedUrl } from '../../../../../api/imageProxy'
+import { ApplicationState } from '../../../../../redux'
+
+export const ImageFrame: React.FC<React.ImgHTMLAttributes<HTMLImageElement>> = ({ alt, src, ...props }) => {
+  const [imageUrl, setImageUrl] = useState(src)
```
mrdrogdrog, line 7: You're using the original src as initial value. That means, that before you get the proxied url, you use the orignal one.

CHANGELOG.md
SISheogorath: I would remove the word "now" here. Keep in mind, this changelog will be around in the 10th future release. Relative time units like now/soon/recently are not overly future proof.

src/api/imageProxy.ts
SISheogorath: Because the image proxy is a separate piece of software that stands besides the backend.

One could implement it in the backend, but then you have to deal with mime-types, images sizes, dead URL, etc. in our backend code and that is just a unneeded hassle. Or we start to pipe things through our backend which is something `node` is not meant to do, due to its single-threaded, event-queue-based system.

mrdrogdrog: That makes sense. I guess it's okay.

mrdrogdrog: Did you think about the possibility to set the proxy url into the frontend config and sending requests directly to the proxy?

SISheogorath: At least camo uses a secret that is shared between a backend and the proxy in order to not being a fully open proxy in order to prevent abuse and apply rate-limiting  on application side. That's at least my idea why the backend should be involved here.

See: https://github.com/cactus/go-camo#how-it-works

mrdrogdrog: Hm okay. How do we prevent that the backend endpoint /media/proxy becomes a ffa proxy?

mrdrogdrog: CSRF?

mrdrogdrog: Or:

Is there any plan to protect this endpoint from abuse with non-hedgedoc use?

mrdrogdrog: Well. I think that doesn't matter here. That's more of a backend story.

src/components/editor/markdown-renderer/replace-components/image/image-frame.tsx
ErikMichelson: Yes, that's a kind of fallback to show an image in any case. It's also the default value for the state as the url-replace-logic won't be triggered at all if the imageProxy setting is not set to true.

CHANGELOG.md
ErikMichelson: Changed that in 4fa8e770. Does it sound better with the change?

src/components/editor/markdown-renderer/replace-components/image/image-frame.tsx
mrdrogdrog: Yes, but your first rendering would ALWAYS set the original src, then the effect would trigger and THEN after all the ajax and callbacks the proxied image would be shown

OR 

The image proxy is useless because the client is loading the original image :sweat_smile:

CHANGELOG.md
mrdrogdrog: I would write something like "Images will be loaded via proxy if an image proxy is configured in the backend".
Because you don't have to mention "The CodiMD frontend". We're already in the frontend :yum:

src/components/editor/markdown-renderer/replace-components/image/image-frame.tsx
mrdrogdrog: I also would NEVER use the original one as fallback, because that could leak data. But we could add a boolean to the backendConfig like "fallbackToOriginalImageUrl". Some people prefer availablity and some privacy.

ErikMichelson: You're right, the fallback makes an image proxy useless. I'll change that.

Regarding the additional config option I'm unsure. If people want CodiMD-content to be highly available, and they're using an image proxy, that proxy should be highly available too.

SISheogorath: Yes, let's not try to be smart here. If one configures an image proxy, use the proxy, if one doesn't then don't. That easy. :+1:

ErikMichelson: Should be done now

CHANGELOG.md
ErikMichelson: Should be done now

src/components/editor/markdown-renderer/replace-components/image/image-frame.tsx
```diff
@@ -1,7 +1,26 @@
...
+  const imageProxyEnabled = useSelector((state: ApplicationState) => state.backendConfig.imageProxy)
+
+  useEffect(() => {
+    if (!imageProxyEnabled) {
+      setImageUrl(src || '')
+      return
+    }
+    // This additional local function is needed because react-useEffect does not allow async functions as effect.
+    const fetchProxiedUrl = async (): Promise<void> => {
+      const proxiedUrlResponse = await getProxiedUrl(src || '')
```
mrdrogdrog, line 17: Why should we start an ajax call for an empty url?

ErikMichelson: Done

public/api/v2/config
```diff
@@ -26,6 +26,7 @@
     "oauth2": "Olaf2",
     "saml": "aufSAMLn.de"
   },
+  "imageProxy": false,
```
mrdrogdrog, line 29: (optional) Maybe "useImageProxy" would be more precise?

src/components/editor/markdown-renderer/replace-components/image/image-frame.tsx
```diff
@@ -1,7 +1,26 @@
...
+    if (!imageProxyEnabled || !src) {
+      setImageUrl(src || '')
+      return
+    }
+    // This additional local function is needed because react-useEffect does not allow async functions as effect.
+    const fetchProxiedUrl = async (): Promise<void> => {
+      const proxiedUrlResponse = await getProxiedUrl(src)
+      setImageUrl(proxiedUrlResponse.src)
+    }
+    fetchProxiedUrl().catch(err => console.error(err))
```
mrdrogdrog, line 20: If you don't like the extra async function.. then don't use it? :grin: 

```suggestion
    getProxiedUrl(src)
      .then(proxiedUrlResponse => setImageUrl(proxiedUrlResponse.src))
      .catch(err => console.error(err))
```

src/components/editor/markdown-renderer/replace-components/image/image-frame.tsx
```diff
@@ -1,7 +1,26 @@
-import React from 'react'
+import React, { useEffect, useState } from 'react'
+import { useSelector } from 'react-redux'
+import { getProxiedUrl } from '../../../../../api/imageProxy'
+import { ApplicationState } from '../../../../../redux'
+
+export const ImageFrame: React.FC<React.ImgHTMLAttributes<HTMLImageElement>> = ({ alt, src, ...props }) => {
```
mrdrogdrog, line 6: What still bugs me about this component is, that it needs two renderings for a non-proxied image. How about rendering the img tag with the src, if the proxy is not enabled?

src/components/editor/markdown-renderer/replace-components/image/image-frame.tsx
ErikMichelson: You're completely right, I was overcomplicating things. :D
Done.

src/components/editor/markdown-renderer/replace-components/image/image-frame.tsx
ErikMichelson: Changed in 581d7028

Issue #324, opened 2020-07-11
PlantUML support


davidmehren:
Adapted from https://github.com/codimd/server/pull/29.

It would be nice to support [PlantUML](https://plantuml.com/) diagrams. It's a bit ugly as you have to do a [bunch of transformations](https://plantuml.com/de/code-javascript-asynchronous) to the PlantUML markup and send that to your PlantUML server (Java) which gives you a PNG or SVG.

**Describe the solution you'd like**
Ability to use PlantUML code like this:
```
```plantuml
(plantuml code)
\```
```
(please ignore the backslash, I couldn't get a codeblock inside a codeblock to work otherwise)

truh:
I got an implementation here https://github.com/truh/react-client/tree/feature/plantuml

mrdrogdrog:
Open a PR and we'll review it.

truh:
How do we want to deal with opt-in rendering on a third party server? Do we need to consider it at this stage or can we take care of it when react-client has a concept of configuration?

mrdrogdrog:
The react client already has a concept of configuration. it requests the frontend configuration, that is needed for bootstraping the client and the backend configuration, that contains all the instance stuff. We could add the plantuml server address to the backend config.

PR #327, opened 2020-07-12
Add plantuml support

truh:
### Component/Part
Add PlantUML rendering support.

### Description
Code block like

    ```plantuml
    Bob -> Alice : hello
    ```
are rendered as PlantUML diagrams using a PlantUML server.

### Steps

- [x] added implementation
- [ ] added / updated tests
- [ ] extended changelog

### Related Issue(s)

#324

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,7 +58,7 @@
     "markdown-it-ins": "3.0.0",
     "markdown-it-mark": "3.0.0",
     "markdown-it-mathjax": "2.0.0",
-    "markdown-it-plantuml": "^1.4.1",
+    "markdown-it-plantuml": "1.4.1",
     "markdown-it-regex": "0.2.0",
     "markdown-it-sub": "1.0.0",
     "markdown-it-sup": "1.0.0",
--- a/package.json
+++ b/package.json
@@ -58,6 +58,7 @@
     "markdown-it-ins": "3.0.0",
     "markdown-it-mark": "3.0.0",
     "markdown-it-mathjax": "2.0.0",
+    "markdown-it-plantuml": "^1.4.1",
     "markdown-it-regex": "0.2.0",
     "markdown-it-sub": "1.0.0",
     "markdown-it-sup": "1.0.0",
--- a/public/api/v2/config
+++ b/public/api/v2/config
@@ -27,6 +27,7 @@
     "saml": "aufSAMLn.de"
   },
   "useImageProxy": false,
+  "plantumlServer": "http://www.plantuml.com/plantuml",
   "specialLinks": {
     "privacy": "https://example.com/privacy",
     "termsOfUse": "https://example.com/termsOfUse",
--- a/public/api/v2/config
+++ b/public/api/v2/config
@@ -27,7 +27,10 @@
     "saml": "aufSAMLn.de"
   },
   "useImageProxy": false,
-  "plantumlServer": "http://www.plantuml.com/plantuml",
+  "plantuml": {
+    "enable": true,
+    "server": "http://www.plantuml.com/plantuml"
+  },
   "specialLinks": {
     "privacy": "https://example.com/privacy",
     "termsOfUse": "https://example.com/termsOfUse",
--- a/public/api/v2/config
+++ b/public/api/v2/config
@@ -27,9 +27,7 @@
     "saml": "aufSAMLn.de"
   },
   "useImageProxy": false,
-  "plantuml": {
-    "enable": true
-  },
+  "plantumlServer": "http://www.plantuml.com/plantuml",
   "specialLinks": {
     "privacy": "https://example.com/privacy",
     "termsOfUse": "https://example.com/termsOfUse",
--- a/public/api/v2/config
+++ b/public/api/v2/config
@@ -28,8 +28,7 @@
   },
   "useImageProxy": false,
   "plantuml": {
-    "enable": true,
-    "server": "http://www.plantuml.com/plantuml"
+    "enable": true
   },
   "specialLinks": {
     "privacy": "https://example.com/privacy",
--- a/src/api/config/types.ts
+++ b/src/api/config/types.ts
@@ -7,6 +7,7 @@ export interface Config {
   useImageProxy: boolean,
   specialLinks: SpecialLinks,
   version: BackendVersion,
+  plantumlServer: string | null,
 }
 
 export interface BrandingConfig {
--- a/src/api/config/types.ts
+++ b/src/api/config/types.ts
@@ -7,10 +7,7 @@ export interface Config {
   useImageProxy: boolean,
   specialLinks: SpecialLinks,
   version: BackendVersion,
-  plantuml: {
-    enable: boolean,
-    server: string | null,
-  },
+  plantumlServer: string | null,
 }
 
 export interface BrandingConfig {
--- a/src/api/config/types.ts
+++ b/src/api/config/types.ts
@@ -7,7 +7,10 @@ export interface Config {
   useImageProxy: boolean,
   specialLinks: SpecialLinks,
   version: BackendVersion,
-  plantumlServer: string | null,
+  plantuml: {
+    enable: boolean,
+    server: string | null,
+  },
 }
 
 export interface BrandingConfig {
--- a/src/components/editor/markdown-renderer/markdown-it-plugins/plantuml-error.ts
+++ b/src/components/editor/markdown-renderer/markdown-it-plugins/plantuml-error.ts
@@ -1,0 +1,19 @@
+import MarkdownIt, { Options } from 'markdown-it/lib'
+import Renderer, { RenderRule } from 'markdown-it/lib/renderer'
+import Token from 'markdown-it/lib/token'
+import './plantuml-error.scss'
+
+export const plantumlError: MarkdownIt.PluginSimple = (md) => {
+  const defaultRenderer: RenderRule = md.renderer.rules.fence || (() => { return '' })
+  md.renderer.rules.fence = (tokens: Token[], idx: number, options: Options, env, slf: Renderer) => {
+    const token = tokens[idx]
+    const defaultRender = defaultRenderer(tokens, idx, options, env, slf)
+    if (token.info === 'plantuml') {
+      return `
+        <p class="plantuml-error">PlantUML plugin is enabled but not properly configured.</p>
+        ${defaultRender}
+      `
+    }
+    return defaultRender
+  }
+}
--- a/src/components/editor/markdown-renderer/markdown-it-plugins/plantuml-error.ts
+++ b/src/components/editor/markdown-renderer/markdown-it-plugins/plantuml-error.ts
@@ -6,15 +6,13 @@ export const plantumlError: MarkdownIt.PluginSimple = (md) => {
   const defaultRenderer: RenderRule = md.renderer.rules.fence || (() => '')
   md.renderer.rules.fence = (tokens: Token[], idx: number, options: Options, env, slf: Renderer) => {
     const token = tokens[idx]
-    const defaultRender = defaultRenderer(tokens, idx, options, env, slf)
     if (token.info === 'plantuml') {
       return `
         <p class="alert alert-danger">
           PlantUML plugin is enabled but not properly configured.
         </p>
-        ${defaultRender}
       `
     }
-    return defaultRender
+    return defaultRenderer(tokens, idx, options, env, slf)
   }
 }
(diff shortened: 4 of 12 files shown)


src/external-types/markdown-it-plantuml/index.d.ts
```diff
@@ -0,0 +1,5 @@
+declare module 'markdown-it-plantuml' {
+  import MarkdownIt from 'markdown-it/lib'
+  const markdownItPlantuml: MarkdownIt.PluginWithOptions
```
ErikMichelson, line 3: It looks as if the options are missing here. Compare with this other markdown-it-plugin for example: [markdown-it-toc-done-right](https://github.com/codimd/react-client/blob/master/src/external-types/markdown-it-toc-done-right/index.d.ts#L3-L4)

truh: done

mrdrogdrog:
@truh Please rebase the branch

package.json
```diff
@@ -58,6 +58,7 @@
     "markdown-it-ins": "3.0.0",
     "markdown-it-mark": "3.0.0",
     "markdown-it-mathjax": "2.0.0",
+    "markdown-it-plantuml": "^1.4.1",
```
mrdrogdrog, line 61: Pin the dependency

src/components/editor/markdown-renderer/markdown-renderer.tsx
```diff
@@ -127,6 +131,13 @@ export const MarkdownRenderer: React.FC<MarkdownRendererProps> = ({ content, onM
       })
     }
     md.use(taskList)
+    if (plantumlServer) {
```
ErikMichelson, line 134: Hm, I'm a bit unsure about this if-statement.

Of course if no PlantUML server is defined, we can't render anything. But maybe we should show an error message in that case.

While for the admin of an instance it's clear that PlantUML is disabled for example, a user that tries to copy a document with a PlantUML diagram from another instance might be confused why it won't work on this instance.

src/components/editor/markdown-renderer/markdown-renderer.scss
```diff
@@ -14,6 +14,11 @@
     max-width: 900px;
   }
 
+  & > img {
+    width: unset;
```
ErikMichelson, line 18: As I'm currently unable to test it (being not at home): Did you test that this statement does not interfere with the wider-images feature in view-only mode?

truh: Seems to be alright. I have also created a PR at the `markdown-it-plantuml` repo to be able to use a class name selector for this. The maintainer hasn't responded so far. https://github.com/gmunguia/markdown-it-plantuml/pull/24

src/components/editor/markdown-renderer/markdown-renderer.tsx
truh: I think that PlantUML should be an optional, opt-in feature. It requires the admin to either setup their own PlantUML server or trust parts of their users documents to a third party server (http://www.plantuml.com/plantuml/).

package.json
truh: done

src/components/editor/markdown-renderer/markdown-renderer.tsx
ErikMichelson: I agree that PlantUML should be opt-in. But I guess it's better anyway to show a message like "PlantUML is not enabled on this server" instead of a silent failure.

DerMolly: The general syntax for plantUML is written between \```plantuml and \```. If the Plugin is not supported the normal rendering for code should be used so I think this is an appropriate  work around.
But it's also true that it's kinda puzzling if plantuml code get's rendered on one server and not on the next. Not sure how that could be reconciled with everything we have right now. An permanent error message above every plantuml code block, letting the user know that their admin did not activate this feature seems rather odd, too.
Maybe the best idea would be (if we extract the plantuml code into a hedgedoc plugin) to show the error when the plugin is included and the plantuml server is not configured, but not if the plugin is absent. I'd suggest opening a separate issue for this and mark this thread as resolved in the mean time.

mrdrogdrog: I don't think that we need to extract the code plantuml code. The markdown parser is initialized on every rendering (something we should probably not do). So if plantuml is activated and possible use the markdown it plugin. Otherwise include a markdown it plugin that replaces the fence with an alert.

src/components/editor/markdown-renderer/markdown-renderer.scss
mrdrogdrog: This should be fine, because the normal images are wrapped in p Tags that have the wider class

src/components/editor/markdown-renderer/markdown-renderer.tsx
mrdrogdrog: Then open an issue and link it here.

mrdrogdrog: Or let's ask @truh. Do you want to add the "plantuml is not activated" banner plugin here or in an extra pr?

public/api/v2/config
```diff
@@ -26,6 +26,7 @@
     "oauth2": "Olaf2",
     "saml": "aufSAMLn.de"
   },
+  "plantumlServer": "http://www.plantuml.com/plantuml",
```
ccoenen, line 29: should this be pre-filled here? (also it's `http`, not `https`)

DerMolly: This is the mock answer to the call to the api, so this is fine for now.

truh: https redirects to http.

src/components/editor/markdown-renderer/markdown-renderer.tsx
truh: If I understand @DerMolly correctly, he doesn't want a banner plugin, he wants this PlantUML code be refactored into a plugin (as soon as a plugin interface is available) and the combination of _plugin is installed_ + _server is not configured_ should show an error message.

mrdrogdrog: If i understand him correcly, he wants to show a message that plantuml code was recognized, but can't be processed because the admin didn't activate plantuml. 
You can implement that easily. You can check if the planuml server url is set and if it is, then you use the plantuml markdown it plugin. If it's not activated, then you use the "show a plantuml is not activate" markdown it plugin.

mrdrogdrog: That's indepentend from a hedgedoc-frontend-plugin-api

truh: Showing a warning in the frontend just sounds like an anti-feature to me. If we are planning for the hedgedoc frontend to have plugin support there will be a lot of cases where rendering differs between instances and having code blocks as a fallback for unsupported languages seems preferable over warning boxes.

mrdrogdrog: I'm not talking about a general fallback for unsupported languages. I'm talking about a warning in this specific case. Something like "Plantuml rendering is installed, but the admin didn't set a plantuml server url. Therefore I can't render your diagram. sorry"

mrdrogdrog: Because then you want plantuml to work, but it doesn't and I guess it's good to let users know WHY it's not working.

truh: I do thing that stuff like this should be handled in a consistent manner.

The reason why something doesn't work should concern the admin of an installation, not the user. If we consider an empty plantumlServer option a configuration error, this should be logged to the server console, not show up on the frontend.

mrdrogdrog: If the user expects this to work and it doesn't work there are two reasons:
The user made a mistake or the admin did it. Why should we confuse the user if we can tell him that it's not his fault? I don't get what's bad with a notice that says "the admin has misconfigured plantuml"

DerMolly: IMHO showing a warning should be included in this PR for now. When we extract this code into a plugin we'll need that anyway and it doesn't hurt to implement it now.

public/api/v2/config
SuperSandro2000: I don't think it is a good idea to use a service which fails to properly implement https and handles user data.

truh: @SuperSandro2000 that's why I don't set the plantumlServer by default. `public/api/v2/config ` is a mock response for the configuration API endpoint, this won't be in production.

mrdrogdrog: I don't even know if we have to include this into the backend config. The plantuml rendering is nothing where the backend is involved. Maybe we should put this into the frontend config instead?

mrdrogdrog: Btw: we don't even know if the plantuml plugin will be included and bundled into the frontend. (After we created the frontend plugin system)

mrdrogdrog: Forget the point with the frontend config. The system with the two configs could be reworked soon.

ccoenen: if this is a mock response, the value should be `https://example.com/plantuml`.

mrdrogdrog: No, because then we don't know if it's working.

truh:
@mrdrogdrog I implemented the warning, took me a while to figure out how.

src/components/editor/markdown-renderer/markdown-it-plugins/plantuml-error.ts
```diff
@@ -0,0 +1,19 @@
...
+import './plantuml-error.scss'
+
+export const plantumlError: MarkdownIt.PluginSimple = (md) => {
+  const defaultRenderer: RenderRule = md.renderer.rules.fence || (() => { return '' })
+  md.renderer.rules.fence = (tokens: Token[], idx: number, options: Options, env, slf: Renderer) => {
+    const token = tokens[idx]
+    const defaultRender = defaultRenderer(tokens, idx, options, env, slf)
+    if (token.info === 'plantuml') {
+      return `
+        <p class="plantuml-error">PlantUML plugin is enabled but not properly configured.</p>
```
mrdrogdrog, line 13: You don't have to define a new alert class. Just use the bootstrap alert.

src/api/config/types.ts
```diff
@@ -7,6 +7,10 @@ export interface Config {
   useImageProxy: boolean,
   specialLinks: SpecialLinks,
   version: BackendVersion,
+  plantuml: {
+    enable: boolean,
```
mrdrogdrog, line 11: I don't think, that we need an extra "enabled"-flag. If the url is set then the plugin is enabled.

src/components/editor/markdown-renderer/markdown-it-plugins/plantuml-error.ts
```diff
@@ -0,0 +1,19 @@
+import MarkdownIt, { Options } from 'markdown-it/lib'
+import Renderer, { RenderRule } from 'markdown-it/lib/renderer'
+import Token from 'markdown-it/lib/token'
+import './plantuml-error.scss'
+
+export const plantumlError: MarkdownIt.PluginSimple = (md) => {
+  const defaultRenderer: RenderRule = md.renderer.rules.fence || (() => { return '' })
```
mrdrogdrog, line 7:
```suggestion
  const defaultRenderer: RenderRule = md.renderer.rules.fence || (() => '')
```

mrdrogdrog:
> @mrdrogdrog I implemented the warning, took me a while to figure out how.

The next time you're stuck, feel free to ask us! We're always happy to help.

src/api/config/types.ts
mrdrogdrog: Ignore this one. We can refactor it, when we extract the code into a plugin.

src/components/editor/markdown-renderer/markdown-it-plugins/plantuml-error.ts
```diff
@@ -0,0 +1,20 @@
...
+  const defaultRenderer: RenderRule = md.renderer.rules.fence || (() => '')
+  md.renderer.rules.fence = (tokens: Token[], idx: number, options: Options, env, slf: Renderer) => {
+    const token = tokens[idx]
+    const defaultRender = defaultRenderer(tokens, idx, options, env, slf)
+    if (token.info === 'plantuml') {
+      return `
+        <p class="alert alert-danger">
+          PlantUML plugin is enabled but not properly configured.
+        </p>
+        ${defaultRender}
```
mrdrogdrog, line 15: Why are you rendering the source code? if plantuml is enabled I wouldn't expect a code block.

DerMolly:
thank you very much @truh

PR #356, opened 2020-07-25
Fix firefox e2e runs

mrdrogdrog:
### Component/Part

E2E tests in headless firefox

### Description
This PR fixes the unstable behaviour of the firefox e2e execution by increasing the time to wait for results when executing a cypress command.

### Steps

- [x] added implementation
- [x] added / updated tests
- [x] extended changelog

### Related Issue(s)
Closes #356

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -16,6 +16,8 @@ jobs:
     steps:
       - uses: actions/checkout@v1
       - uses: cypress-io/github-action@v2
+        env:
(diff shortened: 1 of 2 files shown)


DerMolly:
we could also remove the gcinterval hack in
https://github.com/codimd/react-client/blob/8784da84977335d48fcea8adff9b1a944ce81f8f/cypress.json#L5-L8
as this should be fixed in 4.11.0, which we now use see https://github.com/cypress-io/cypress/issues/6813

PR #361, opened 2020-07-26
rework how the frontend is started

### LINKED ISSUE #358: Merge backend and frontend config
The frontend config

Right now we have two kind of config files. The frontend and the backend config. The frontend config is placed into the public folder of the frontend and tells the frontend where it can find the backend. The backend config tells the frontend all the details about auth, branding, etc.

But if you think about it, then you come to the conclusion that it's useless to do two ajax calls on the same domain. Especially because the primary deployment procedure is, that the backend delivers the frontend. 

We merge both config files and let the backend generate the config on the fly with all the keys.
For development, if you use the frontend dev server and the backend dev server on two separate ports, we can use an environment variable to tell the frontend where it can find the config file. If the variable is set, the frontend dev server uses this URL to call the config. If not, it falls back to the url that is set now.

DerMolly:
### Component/Part
The whole app

### Description
This PR changes how the frontend runs to better suite it for the future development side-by-side withe the backend and the production stage, where it will be served by the backend.
Specifically the following was done: 
- renamed frontend-config to api-url
- renamed backend-config to config
- removed api call to set frontend-config as the frontend either know where the backend is as it is delivered by it or get's this information via the enviroment variable `REACT_APP_BACKEND`
- always start the client on port `3001` as the backend will run on `3000` during development. This is a good default for the backend,as this will be the port on which it will run on most users machines.
- added package.json script 'start:dev', which will set `REACT_APP_BACKEND` to `http://localhost:3000`
- changed README to better explain how to run backend and frontend side-by-side

### Steps

- [x] added implementation
- [x] added / updated tests
- [ ] extended changelog

### Related Issue(s)
closes #358


src/api/config/index.ts
```diff
@@ -0,0 +1,12 @@
+import { expectResponseCode, getApiUrl } from '../../utils/apiUtils'
+import { defaultFetchConfig } from '../default'
+import { Config } from './types'
+
+export const getConfig = async (): Promise<Config> => {
+  console.debug(getApiUrl())
```
mrdrogdrog, line 6: Debug output

src/components/application-loader/application-loader.tsx
```diff
@@ -9,8 +9,7 @@ export const ApplicationLoader: React.FC = ({ children }) => {
   const { pathname } = useLocation()
 
   const setUpTasks = useCallback(() => {
-    const baseUrl: string = window.location.pathname.replace(pathname, '') + '/'
-    console.debug('Base URL is', baseUrl)
+    const baseUrl: string = window.location.protocol + '//' + window.location.host.replace(pathname, '')
```
mrdrogdrog, line 12: Keep the old method. Otherwise we can't run the client in a subfolder

src/components/application-loader/initializers/configLoader.ts
```diff
@@ -1,24 +1,21 @@
...
-  setFrontendConfig(frontendConfig)
+  setApiUrl({
+    apiUrl: (process.env.REACT_APP_BACKEND || baseUrl) + '/api/v2'
+  })
 
-  const backendConfig = await getBackendConfig()
-  if (!backendConfig) {
+  const config = await getConfig()
+  if (!config) {
     return Promise.reject(new Error('Backend config empty!'))
```
mrdrogdrog, line 14: Reference to 'backend'

src/api/config/index.ts
DerMolly: removed

src/components/application-loader/application-loader.tsx
DerMolly: did that

src/components/application-loader/initializers/configLoader.ts
DerMolly: done

cypress/support/config.ts
```diff
@@ -5,7 +5,7 @@ export const banner = {
 
 export const branding = {
   name: 'ACME Corp',
-  logo: 'http://localhost:3000/acme.png'
+  logo: 'http://localhost:3001/acme.png'
```
mrdrogdrog, line 8:
```suggestion
  logo: '/acme.png'
```

Issue #93, opened 2020-05-30
Reordering ui elements for a clear information architecture


ccoenen:
A lot of the UI elements in CodiMD were thrown in "where there was some space left", or so it feels. I would love to see this rewrite deal with this.

Basically I suggest to have two clear hierarchical zones:

The **top bar** (needs a better name) should be reserved for everything that is of overarching interest in CodiMD

Below that the **document area** (probably also needs a better name) would start with _everything related to the document_. There are some elements that need to be in the document area regardless of edit mode _or_ view mode. So I'm essentially proposing a second bar of UI elements with everything related to the document in between the current main bar and the view/edit zone.

This is a legacy CodiMD Screenshot, basically I think these things should change:
* Keep the **logo** (1)
* Deal with the **mode switcher** (2) once we've discussed #92 (it is then either a global thing and stays there _or_ it is a document thing and moves down a bit).
* **Night mode** (3) is global as far as I know, so it stays there.
* **Help** (4) is global, so it stays there.
* **New Document** (5) is global, so it stays there.
* "**x people online**" (6) is a document level info, this should move to the document area (i.e. Down).
* in its place (7) there should be info on the **currently logged in user** (or the possibility to log in if one is _not yet_ logged in)
* **Publish** (8) should be incorporated into (9) and both belong to the document (i.e. they should move down a bit)
* As mentioned earlier the "**menu**" (9) is only related to stuff about the document, so it should probably *get a different name*. And it needs to move down to the document tools.
* The **permissions** (10) are currently only visible in "view" mode, and I would like to change this. This is document level stuff that belongs with (6), (8) and (9), as such it would be visible regardless of edit or view or both modes. As mentioned in #91, 'delete' should not be in that one but in (9) instead, in my opinion.
* **meta info** on ownership and changes (11) may also belong int that bar, I'm not sure about those.
* (not pictured) the **in-document navigation sidebar** should become part of the document level bar so it is available in edit mode as well

Numbers are color-coded for "general level" (fuchsia), "document level" (turquois) and "undecided" (blue).

![grafik](https://user-images.githubusercontent.com/124909/83328660-f64f5b00-a284-11ea-9689-12a8a807587a.png)

mrdrogdrog:
Solved in #399

PR #399, opened 2020-08-08
Restructure editor user interface

### LINKED ISSUE #331: Add 'pin'/'favorite' button in the note editor
Currently, it is possible to 'pin' a note in the note history view.
This feature should also be available in the note editor itself.
It seems counter-intuitive to read a note, find it 'pin worthy' and then need to switch to the history view to actually pin it.

The easiest way to add this is probably to add a pin (or favorite star) button in the top taskbar.

This is blocked by #330

ErikMichelson:
### Component/Part

Editor -> UI

### Description
This PR restructures the user interface of HedgeDoc's heart component: The editor.
To separate concerns of the application and the opened document itself, a new document-bar is introduced directly below the top-bar which only contains application-specific actions from now on. The information about note creation and last edits moved into a new modal dialog "document info" that can be invoked from the document bar. The toolbar of the editor is now grouped into different sections and the settings from the status-bar of the editor will move into a new editor-settings-modal invoked from the toolbar as well.

Note that this interface is still a draft and not settled into stone yet. Feel free to add comments to this PR below.

image

image

image

### Steps

- [ ] added implementation
- [ ] added / updated tests
- [ ] extended changelog

### Related Issue(s)
closes #331 
see #93


ccoenen:
One thing I'm not clear about is, why in the second menu bar some things are left-aligned with icons and some things are right-aligned. And why is import/export/share three buttons in two places?

I would at least request that the permissions thing gets moved to the right side (where all the dropdowns are) directly next to the "X online" dropdown. These two are the only things that are dropdowns _AND_ display a state (as opposed to a label like import/export).

Another thing I'd like to ask is, what is in the "menu" dropdown? The things that _used to be_ there are now directly in the menubar, are they not?

mrdrogdrog:
> One thing I'm not clear about is, why in the second menu bar some things are left-aligned with icons and some things are right-aligned.

It looked a bit odd, when everything was on the right, so we split the actions from the menus. 

> And why is import/export/share three buttons in two places?

We have enough space. Why shouldn't we use it?

Share is not export. If that isn't clear enough, then maybe we should rename it to "share link"?

> I would at least request that the permissions thing gets moved to the right side (where all the dropdowns are) directly next to the "X online" dropdown. These two are the only things that are dropdowns _AND_ display a state (as opposed to a label like import/export).

Good point.

> Another thing I'd like to ask is, what is in the "menu" dropdown? The things that _used to be_ there are now directly in the menubar, are they not?

In the menu "menu" is now only "delete note". We still don't know where to put this action and wanted to keep the menu until the PR is ready to review. If we don't have any good reason for keeping it, then we'll delete the menu and put the action somewhere else.

My idea was to rename this menu to "delete note", but keep the menu item. Because then you'll have to click a dropdown menu, the dropdown icon AND confirm a modal. IMHO you can't have enough clicks for such a destructing action.

DerMolly:
Also regarding the menu menu, maybe we find other items that don't fit anything else.

mrdrogdrog:
We changed the permission dropdown to a button, that opens a modal, because the permission system is under construction and will be changed in the future.

ccoenen:
thanks :-)

public/locales/en.json
```diff
@@ -219,18 +219,44 @@
...
+            "pinnedToHistory": "Pinned to history"
+        },
+        "preferences": {
+            "title": "Preferences",
+            "theme": "Theme",
+            "keymap": "Keymap",
+            "tabCharacter": "Tab character",
+            "tabSize": "Tab size",
+            "spellChecker": "Spell checking"
+        },
```
mrdrogdrog, line 250: shouldn't that be under "modal" ?

src/components/editor/editor-window/tool-bar/editor-preferences/editor-preferences.tsx
```diff
@@ -0,0 +1,67 @@
...
+import { EditorPreferenceTabsize } from './editor-preference-tabsize'
+import { EditorPreferenceTheme } from './editor-preference-theme'
+
+export interface EditorSettingsButtonProps {
+  editor: CodeMirror.Editor
+  preferences: EditorConfiguration
+  onPreferencesChange: (config: EditorConfiguration) => void
+}
+
+export const EditorPreferences: React.FC<EditorSettingsButtonProps> = ({ editor, onPreferencesChange, preferences }) => {
```
mrdrogdrog, line 18: I don't know if it's a good idea to create 4 preference specific components, that all have similar code. They all are just a dropdown for text. I think a more elegant solution would be to create a generic "select dropdown with label" component and maintain their options and the state manipulation here.

src/components/landing/layout/user-avatar/user-avatar.tsx
```diff
@@ -1,13 +1,15 @@
...
 export interface UserAvatarProps {
     name: string;
     photo: string;
     additionalClasses?: string;
+    showName?: boolean
 }
 
-const UserAvatar: React.FC<UserAvatarProps> = ({ name, photo, additionalClasses = '' }) => {
+const UserAvatar: React.FC<UserAvatarProps> = ({ name, photo, additionalClasses = '', showName = true }) => {
   // ToDo: add Translation Key for Avatar of ${name}
```
mrdrogdrog, line 13: Can you solve this todo? :sweat_smile:

src/components/editor/editor-window/tool-bar/editor-preferences/editor-preferences.tsx
ErikMichelson: This was a work-in-progress state and now finally just uses a single preference-selection component.

public/locales/en.json
ErikMichelson: The translation keys were restructured. "preferences" is now under "modal".

src/components/landing/layout/user-avatar/user-avatar.tsx
ErikMichelson: Done in 3b462281

ccoenen:
I'd like to request an updated screenshot, if that is possible.

ErikMichelson:
@ccoenen Feel free to play around with it here: https://hedgedoc-ui.test.fsbremen.de/n/features

mrdrogdrog:
@ccoenen 

image
image
image

PR #432, opened 2020-08-18
Use prod build for e2e tests

mrdrogdrog:
### Component/Part
The e2e tests

### Description
This PR changes the e2e workflow. It now uses the prod build, because that is what we actually want to test. It could also fix the firefox timeout.

### Steps

- [x] added implementation
- [x] added / updated tests
- [ ] extended changelog

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -2893,6 +2893,11 @@ base@^0.11.1:
     mixin-deep "^1.2.0"
     pascalcase "^0.1.1"
 
+basic-auth@^1.0.3:
+  version "1.1.0"
+  resolved "https://registry.yarnpkg.com/basic-auth/-/basic-auth-1.1.0.tgz#45221ee429f7ee1e5035be3f51533f1cdfd29884"
+  integrity sha1-RSIe5Cn37h5QNb4/UVM/HN/SmIQ=
+
 batch@0.6.1:
   version "0.6.1"
   resolved "https://registry.yarnpkg.com/batch/-/batch-0.6.1.tgz#dc34314f4e679318093fc760272525f94bf25c16"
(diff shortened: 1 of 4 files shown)


yarn.lock
```diff
@@ -1799,7 +1799,7 @@
   dependencies:
     "@types/markdown-it" "*"
 
-"@types/markdown-it@10.0.2":
+"@types/markdown-it@*", "@types/markdown-it@10.0.2":
```
DerMolly, line 1802: did we want to avoid * imports

yarn.lock
```diff
@@ -4186,7 +4196,7 @@ cyclist@^1.0.1:
   resolved "https://registry.yarnpkg.com/cyclist/-/cyclist-1.0.1.tgz#596e9698fd0c80e12038c2b82d6eb1b35b6224d9"
   integrity sha1-WW6WmP0MgOEgOMK4LW6xs1tiJNk=
 
-cypress@4.12.1:
+cypress@*, cypress@4.12.1:
```
DerMolly, line 4199: same here

DerMolly (changes requested):

yarn.lock
mrdrogdrog: We want to avoid separated references. Combined ones are okay and work as expected.

ErikMichelson (approved):
Seems to work fine, although I'm interested about where these requests (typing h-t-t-p-s in the URL bar) originate from?

image

mrdrogdrog:
> Seems to work fine, although I'm interested about where these requests (typing h-t-t-p-s in the URL bar) originate from?
> 
> image

I saw that too.. And I have NO idea o.ö

DerMolly (approved):

yarn.lock
mrdrogdrog: same here

PR #485, opened 2020-08-26
Add revisions dialog

### LINKED ISSUE #425: Revision Screen
The current revision screen looks like this 
grafik
We need to implement something like this for this client to.
Currently we only show the timestamp of the revision and it's length as this are the only information in a `NoteRevisionsMetadata` object (see https://github.com/codimd/server/blob/58e6d91286e8a4ea36c75990f44940ab2610caa3/docs/dev/public_api.yml#L686-L700).
Maybe we could also show which users did work on that revision? This would need a change in the way the API works I guess (what do you think @codimd/developers?)

Also as shown in th scrrenshot the ability to revert to a specific revision and to download the note at a specific point in time should be possible. As one can see in `NoteRevision`s (see https://github.com/codimd/server/blob/58e6d91286e8a4ea36c75990f44940ab2610caa3/docs/dev/public_api.yml#L701-L717), we can get the revisions content and a patch from the last revision to the one we selected, so this should be easy enough.

ErikMichelson:
### Component/Part
Editor -> Document bar -> Revisions

### Description
This PR adds the modal to view and download former revisions of the note.

Please note that this PR resolves the revision list only for the 'features' note and only by using a non-API-confirming call. This is owed to the fact that our mock-up is file-based and there can't be a folder and a file with the same name. A fix for this will be implemented as soon as we have a dynamic backend to test with.

### Steps

- [x] added implementation
- [ ] added / updated tests
- [ ] extended changelog

### Additional information
image

### Related Issue(s)
closes #425


DerMolly:
Maybe the modal should show the top most revision as default? At least this is the way 1.6 does it currently

src/components/editor/document-bar/revisions/revision-modal.tsx
```diff
@@ -0,0 +1,101 @@
...
+    }).catch(() => setError(true))
+  }, [setRevisions, setError, id])
+
+  useEffect(() => {
+    if (selected === null) {
+      return
+    }
+    getRevision(id, selected).then(fetchedRevision => {
+      setSelectedRevision(fetchedRevision)
+    }).catch(() => setError(true))
```
DerMolly, line 41: should old selected revisions be cached?

DerMolly (changes requested):

src/components/editor/document-bar/revisions/revision-modal.tsx
ErikMichelson: Done in 18c1cab

ErikMichelson:
> Maybe the modal should show the top most revision as default? At least this is the way 1.6 does it currently

Done in f490d85

DerMolly (approved):

mrdrogdrog:
The yarn lock is outdated

mrdrogdrog (approved):

PR #493, opened 2020-08-26
Fix editor being filled with €-symbols when pressing the editor-mode-shortcut on certain platforms

ErikMichelson:
### Component/Part
Editor -> Shortcuts

### Description
This fix needs a little bit of explanation. Normally the 'keyUp' prevent is preferred because it fires after being sure, that not some buttons more belong to the shortcut. Additionally, some platforms set the key-property of the 'keyDown'-event to the composition result of the key-combination thus resolving under certain environments to '€' while in other environments to 'e'.
The browser's event flow is as following: keyDown -> keyPress -> textInput -> keyUp.
As the keyUp-event is too late (after textinput) and the keyPress-event does not work properly with the modifiers, we felt compelled to use 'keyDown' and watch for 'e' as well as '€' key-properties. If some other keyboard locale does output different characters than these two, that person got a problem - meaning no functionality of the shortcut. But still better than nothing.

### Steps

- [x] added implementation
- [ ] added / updated tests
- [ ] extended changelog

### Related Issue(s)
none

### DIFF
--- a/src/components/editor/editor.tsx
+++ b/src/components/editor/editor.tsx
@@ -66,9 +66,9 @@ export const Editor: React.FC = () => {
   }, [updateDocumentTitle])
 
   useEffect(() => {
-    document.addEventListener('keyup', shortcutHandler, false)
+    document.addEventListener('keydown', shortcutHandler, false)
     return () => {
-      document.removeEventListener('keyup', shortcutHandler, false)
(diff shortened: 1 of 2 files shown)


mrdrogdrog (approved):
Hacky.. but okay

DerMolly (approved):

PR #511, opened 2020-08-29
Improve render performance

### LINKED ISSUE #359: Improve renderperformance
**Description**
If you add a new paragraph to any markdown content, then too many html nodes will be updated.

**To Reproduce**
1. Create a markdown note with 3 paragraphs. 
2. Add a new paragraph in the markdown before the existing
3. All paragraphs in the html will be updated.

**Expected behavior**
A new html node will be inserted

**Additional context**
This also causes, that special components like videos will be reseted to the "click to load" state and images will be reloaded.

mrdrogdrog:
### Component/Part
Markdown renderer

### Description
This PR:
- Replaces the codimd-linemarker tags with an in-memory-map of positions
- adds persistent keys, that are tied to the source markdown line to the dom elements in the renderer.

### Steps

- [x] added implementation
- [ ] added / updated tests
- [ ] extended changelog

### Related Issue(s)
Closes #359


src/components/markdown-renderer/replace-components/quote-options/quote-options-replacer.tsx
```diff
@@ -37,6 +37,6 @@ export class QuoteOptionsReplacer implements ComponentReplacer {
       return
     }
     node.attribs = Object.assign(node.attribs || {}, { style: `border-left-color: ${attributes['data-color']};` })
-    return subNodeConverter(node, index)
+    return undefined
```
DerMolly, line 40: don't we need that?

mrdrogdrog: No. A prior version had a bug, where the execution of the sub node converter started a loop and a stack overflow. I fixed this.

DerMolly (approved):

src/components/markdown-renderer/renderer-utils.ts
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  delete node.attribs['data-key']
+  return convertNodeToElement(node, key, transform)
+}
+
+export const buildTransformer = (lineKeys: (LineKeys[] | undefined), allReplacers: ComponentReplacer[]):Transform => {
+  const transform: Transform = (node, index) => {
+    const subNodeConverter: SubNodeConverter = (subNode: DomElement, subIndex: number) => {
+      const key = calculateKeyFromLineMarker(subNode, lineKeys) ?? index * 1000000 + subIndex
```
ErikMichelson, line 92: Magic number?

src/components/markdown-renderer/renderer-utils.ts
```diff
@@ -0,0 +1,105 @@
...
+      return renderNativeNode(subNode, key, transform)
+    }
+
+    const key = calculateKeyFromLineMarker(node, lineKeys) ?? index + 2000000
```
ErikMichelson, line 96: Magic number?

ErikMichelson (changes requested):
There are some rare scroll-jumping bugs still present, but more importantly this breaks the KaTeX-block-rendering (inline is fine).

Additional remarks: If we opt to use uuids in the diagram-renderers, why not use uuids here as keys too instead of this offset approach?

ErikMichelson:
KaTeX seems to work now, but inserting new yaml-metadata properties results in a crash.

SISheogorath:
As this tries to improve performance we might want to do a simple benchmark what difference it actually makes. Before and after, just to avoid that things actually get slower by accident.

Just in case it's not known how to do that: https://stackoverflow.com/a/1975103

We might need to adjust things for react a bit and benchmark more individual functions, but generally speaking it sounds doable without great effort but with some useful insight.

ErikMichelson:
Regarding benchmarks: We didn't do a real benchmark yet but tested it out with a ~100.000 chars document (max-length) and compared the speed of scrolling and inserting/modifying of characters between version 1.6 and this branch.

The input-delay of inserting and modifying a character is roughly the same as with 1.6. Fast scrolling a huge document in 1.6 results in the codemirror-editor-pane to be blank and load the text after a few hundred milliseconds. It seems that this version here doesn't have that effect but shows the text instantly while scrolling.

Besides the already mentioned `console.time` and `performance.now` measurement-methods, it could be worth having a look at the React Profiler API: https://reactjs.org/docs/profiler.html.

mrdrogdrog:
@ErikMichelson

mrdrogdrog:
@ErikMichelson 

> Additional remarks: If we opt to use uuids in the diagram-renderers, why not use uuids here as keys too instead of this offset approach?

No, because ReactHtmlParser wants numbers as keys.

src/components/markdown-renderer/markdown-renderer.tsx
```diff
@@ -120,27 +121,49 @@ export const MarkdownRenderer: React.FC<MarkdownRendererProps> = ({ content, onM
...
-      if (!equal(lineMarkerPositions, lastLineMarkerPositions.current)) {
-        lastLineMarkerPositions.current = lineMarkerPositions
-        onLineMarkerPositionChanged(lineMarkerPositions)
+    if (!(documentElement.current && onLineMarkerPositionChanged)) {
+      return
+    }
+    if (lineMarkersMap.current === undefined) {
+      return
+    }
+    const lineMarkersMap_ = lineMarkersMap.current
```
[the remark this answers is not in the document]
mrdrogdrog: done

src/components/markdown-renderer/renderer-utils.ts
mrdrogdrog: Extracted into const

src/components/markdown-renderer/renderer-utils.ts
mrdrogdrog: removed

src/components/markdown-renderer/replace-components/possible-wider/possible-wider-replacer.scss
```diff
@@ -0,0 +1,8 @@
+.wider  .markdown-body {
```
ErikMichelson, line 1:
```suggestion
.wider .markdown-body {
```
Whitespace

src/components/markdown-renderer/replace-components/possible-wider/possible-wider-replacer.tsx
```diff
@@ -1,26 +1,29 @@
...
-    const childIsPDF = node.children[0].name === 'codimd-pdf'
-    if (!(childIsImage || childIsYoutube || childIsVimeo || childIsAsciinema || childIsPDF)) {
+    if (!node.children || node.children.length === 0) {
       return
     }
 
-    // This appends the 'wider-possible' class to the node for a wider view in view-mode
-    node.attribs = Object.assign(node.attribs || {}, { class: `wider-possible ${node.attribs?.class || ''}` })
-    return subNodeConverter(node, index)
+    if (node.children.find((subNode) => subNode.name && enabledTags.indexOf(subNode.name) > -1)) {
```
ErikMichelson, line 21:
```suggestion
    if (node.children.find((subNode) => subNode.name && enabledTags.includes(subNode.name))) {
```

src/components/markdown-renderer/replace-components/possible-wider/possible-wider-replacer.tsx
```diff
@@ -1,26 +1,29 @@
...
+      if (!node.attribs) {
+        node.attribs = {}
+      }
+
+      node.attribs.class += 'wider-possible'
```
ErikMichelson, line 26: This results in two problems.

First: There's a missing space so that the `wider-possible` will be concatenated with the previous class name resulting in non-working classes.

Second: If the class was not set in beforehand, it's value is `undefined` resulting in a total class name like `undefinedwider-possible`.

ErikMichelson (changes requested):

ErikMichelson (approved):

PR #521, opened 2020-08-31
added title attribute to images

### LINKED ISSUE #518: Add support for title-attribute-syntax in images
The following input `alt text` should render as an img-tag with included title-attribute set to "example image".

Dropping this 1.6 feature. But why should we?

DerMolly:
### Component/Part
markdown renderer

### Description
This PR adds the title attribute to images.

### Steps

- [x] added implementation
- [ ] added / updated tests
- [ ] extended changelog

### Related Issue(s)
fixes #518

### DIFF
--- a/src/components/markdown-renderer/replace-components/image/image-frame.tsx
+++ b/src/components/markdown-renderer/replace-components/image/image-frame.tsx
@@ -3,7 +3,7 @@ import { useSelector } from 'react-redux'
 import { getProxiedUrl } from '../../../../api/media'
 import { ApplicationState } from '../../../../redux'
 
-export const ImageFrame: React.FC<React.ImgHTMLAttributes<HTMLImageElement>> = ({ alt, src, ...props }) => {
+export const ImageFrame: React.FC<React.ImgHTMLAttributes<HTMLImageElement>> = ({ alt, src, title, ...props }) => {
   const [imageUrl, setImageUrl] = useState('')
   const imageProxyEnabled = useSelector((state: ApplicationState) => state.config.useImageProxy)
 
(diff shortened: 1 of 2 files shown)


src/components/markdown-renderer/replace-components/image/image-frame.tsx
```diff
@@ -3,7 +3,7 @@ import { useSelector } from 'react-redux'
 import { getProxiedUrl } from '../../../../api/media'
 import { ApplicationState } from '../../../../redux'
 
-export const ImageFrame: React.FC<React.ImgHTMLAttributes<HTMLImageElement>> = ({ alt, src, ...props }) => {
+export const ImageFrame: React.FC<React.ImgHTMLAttributes<HTMLImageElement>> = ({ alt, src, title, ...props }) => {
```
mrdrogdrog, line 6: Is this necessary? The title prop is covered by `...props`

mrdrogdrog (changes requested):

src/components/markdown-renderer/replace-components/image/image-frame.tsx
DerMolly: The same is true for `alt`? I wanted to make it obvious that the title get's set.
To answer the questions: No, it is not.

mrdrogdrog: The alt tag is explicitly set, because otherwise the linter is screaming, because it should be set.

mrdrogdrog: I get your point. If you want to set the title explicitly, too, then it's fine for me.

DerMolly: then would you resolve this an review again?

mrdrogdrog (approved):

ErikMichelson (approved):

PR #527, opened 2020-09-01
Replace editor toolbar dividers with css pseudo-selector

ErikMichelson:
### Component/Part
Editor -> Toolbar

### Description
The use of extra divider elements is not needed as we can set the css-after pseudo-class to the button groups with the same styling.
This way we can reduce the amount of elements in the DOM by a hand full.

### Steps

- [x] added implementation
- [ ] added / updated tests _(not necessary)_
- [ ] added / updated documentation _(not necessary)_
- [ ] extended changelog _(not necessary)_

### Related Issue(s)
none

### DIFF
--- a/src/components/editor/editor-pane/tool-bar/tool-bar.scss
+++ b/src/components/editor/editor-pane/tool-bar/tool-bar.scss
@@ -1,11 +1,11 @@
 .btn-toolbar {
   border: 1px solid #ededed;
-}
 
-.btn-toolbar .btn-group:not(:last-of-type)::after {
-  background-color: #e2e6ea;
-  width: 2px;
-  padding: 0.25rem 0;
-  content: ' ';
-  margin-left: 0.5rem;
+  .btn-group:not(:last-of-type)::after {
+    background-color: #e2e6ea;
+    width: 2px;
+    padding: 0.25rem 0;
+    content: ' ';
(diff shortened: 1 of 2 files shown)


src/components/editor/editor-pane/tool-bar/tool-bar.scss
```diff
@@ -2,8 +2,10 @@
   border: 1px solid #ededed;
 }
 
-.divider {
+.btn-toolbar .btn-group:not(:last-of-type)::after {
```
ccoenen, line 5: `:not(:last-of-type)::after` nice!

DerMolly (approved):

src/components/editor/editor-pane/tool-bar/tool-bar.scss
mrdrogdrog, line 5: Please move this code block into the rule above as sub rule.

mrdrogdrog (changes requested):

src/components/editor/editor-pane/tool-bar/tool-bar.scss
ErikMichelson: Done

mrdrogdrog (approved):
:ok_hand:

PR #532, opened 2020-09-03
build new permissions modal

### LINKED ISSUE #424: New Permissons screen
According to https://github.com/codimd/server/blob/58e6d91286e8a4ea36c75990f44940ab2610caa3/docs/dev/public_api.yml the note permissions will look something like this

```typescript
interface NotePermissions {
  owner: string
  sharedTo: {
    username: string
    canEdit: boolean
  }[]
}
```

We need to implement an UI for these new permissions. Some thoughts on that:
- We'll most likely show the whole thing either as a flotable object (like the emoji picker) or as a modal
- at the top the owner should be shown. <del>If the current user is the owner, they should be able to give away the ownership of the note (this is intended right @codimd/developers?)</del>
- we should show each entry in `sharedTo` in a line (most likely using `UserAvatar`) with a checkbox for `canEdit`. There should also be an `X` for removing the entry. I think depending on if this is an instant action or needs to be saved later (see further down) we can forego an "are you really sure" dialog here, I think.
- underneath should be an input field here one can enter a username to add to the list.
  - we don't want to leak infos about other users on the system, so no autocompletion on the names.
  - but how will we determine if a username is correct or not? Or do we add anything the user wants and simply let the backend decide what legit? How could we then prevent gaining the information with a lot of requests?
- another question we need to think about: Should the dialog be saving on each change or should there be an dedicated save button at the bottom?

DerMolly:
### Component/Part
permissions screen

### Description
This PR adds a new permission screen

### Steps

- [x] added implementation
  - [x] added owner
  - [x] added user list
  - [x] added group list
- [ ] added / updated tests
- [ ] extended changelog

### Related Issue(s)
fixes #424


DerMolly:
This should probably include some e2e tests…

ErikMichelson (changes requested):
All in all it looks good. Just some minor things to fix or clarify.

src/components/editor/document-bar/permissions/permission-modal.tsx
```diff
@@ -0,0 +1,140 @@
...
+
+  const changeGroupMode = (groupId: Principal['id'], canEdit: Principal['canEdit']) => {
+    setGroupList(list =>
+      list
+        .map(group => {
+          if (group.id !== groupId) {
+            return group
+          }
+          group.canEdit = canEdit
+          return group
```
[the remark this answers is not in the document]
DerMolly: done

src/components/editor/document-bar/permissions/permission-modal.tsx
```diff
@@ -0,0 +1,140 @@
...
+
+  const changeUserMode = (userId: Principal['id'], canEdit: Principal['canEdit']) => {
+    setUserList(list =>
+      list
+        .map(user => {
+          if (user.id !== userId) {
+            return user
+          }
+          user.canEdit = canEdit
+          return user
```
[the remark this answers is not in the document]
DerMolly: done

public/locales/en.json
```diff
@@ -320,7 +320,18 @@
                 "button": "Delete note"
             },
             "permissions": {
-                "title": "Permissions"
+                "title": "Permissions",
+                "owner": "Owner",
+                "sharedWithUsers": "Shared with Users",
+                "sharedWithGroups": "Shared with Groups",
```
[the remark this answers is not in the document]
DerMolly: done

src/components/editor/document-bar/permissions/permission-button.tsx
```diff
@@ -0,0 +1,14 @@
+import React, { Fragment, useState } from 'react'
+import { TranslatedIconButton } from '../../../common/icon-button/translated-icon-button'
+import { PermissionModal } from './permission-modal'
+
+export const PermissionButton: React.FC = () => {
+  const [showReadOnly, setShowReadOnly] = useState(false)
```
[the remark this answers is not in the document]
DerMolly: changed

src/components/editor/document-bar/permissions/permission-modal.tsx
```diff
@@ -52,9 +52,8 @@ export const PermissionModal: React.FC<PermissionsModalProps> = ({ show, onChang
       list
         .map(user => {
           if (user.id !== userId) {
```
ErikMichelson, line 54: Actually this needs to be refactored too.
```suggestion
          if (user.id === userId) {
```

ErikMichelson (changes requested):

ErikMichelson (approved):

src/components/editor/document-bar/permissions/permission-modal.tsx
```diff
@@ -0,0 +1,138 @@
...
+    id: 'developer',
+    photo: '',
+    name: 'Developer',
+    canEdit: true
+  }, {
+    id: 'frontend',
+    photo: '',
+    name: 'Frontend',
+    canEdit: true
+  }])
```
mrdrogdrog, line 48: Shouldn't the user data be retreived via API calls?

mrdrogdrog (changes requested):

src/components/editor/document-bar/permissions/permission-modal.tsx
DerMolly: It could be retrieved either by the websocket or an http api in the future, but as we need to know when this changes immediately I guess the only logical solution would be websocket. As we currently don't mock a websocket, I just used static mock data instead.

mrdrogdrog: Why should that come via websocket? The websocket is only for live data.

DerMolly: And you would classify the information that you just lost access to the note or can edit now as unlicensed data?

mrdrogdrog: I'm talking about the user information. Not the permission list itself.

mrdrogdrog: Name, Avatar, etc.

DerMolly: Fair Point, das lässt sich einrichten.

src/components/editor/document-bar/permissions/permission-list.tsx
```diff
@@ -0,0 +1,98 @@
...
+              value={newEntry}
+              placeholder={t(addI18nKey)}
+              aria-label={t(addI18nKey)}
+              onChange={event => setNewEntry(event.currentTarget.value)}
+            />
+            <Button
+              variant='light'
+              className={'text-secondary ml-2'}
+              title={t(addI18nKey)}
+              onClick={() => addEntry()}
```
ErikMichelson, line 89: Is there any reason for this being a separate anonymous function?

```suggestion
              onClick={addEntry}
```

src/components/editor/document-bar/permissions/permission-modal.tsx
```diff
@@ -0,0 +1,162 @@
...
+          name: response.name,
+          photo: response.photo,
+          canEdit: shareUser.canEdit
+        }]))
+      }).catch(() => setError(true))
+    })
+    // set group List
+    permissionsApiResponse.sharedToGroup.forEach(sharedGroup => {
+      if (sharedGroup.id === '1') {
+        console.log('found everyone')
```
ErikMichelson, line 83: The console.log should probably be removed.

src/components/editor/document-bar/permissions/permission-modal.tsx
```diff
@@ -0,0 +1,162 @@
...
+        setAllPermissions(sharedGroup.canEdit ? GroupMode.EDIT : GroupMode.VIEW)
+      } else if (sharedGroup.id === '2') {
+        console.log('found everyone logged in')
```
ErikMichelson, line 86: The console.log should probably be removed.

src/components/editor/document-bar/permissions/permission-modal.tsx
```diff
@@ -0,0 +1,162 @@
...
+  }]
+}
+
+export const PermissionModal: React.FC<PermissionsModalProps> = ({ show, onChangeShow }) => {
+  useTranslation()
+  const [error, setError] = useState(false)
+  const [userList, setUserList] = useState<Principal[]>([])
+  const [owner, setOwner] = useState<UserAvatarProps>()
+  const [allPermissions, setAllPermissions] = useState(GroupMode.NONE)
+  const [loggedPermissions, setLoggedPermissions] = useState(GroupMode.NONE)
```
ErikMichelson, line 59: Maybe it's just me, but I don't like these names. In my opinion they should indicate that they belong to these special groups "allUsers"/"everyone" and "loggedInUsers".

public/locales/en.json
```diff
@@ -320,7 +320,21 @@
...
             },
             "permissions": {
-                "title": "Permissions"
+                "title": "Permissions",
+                "owner": "Owner",
+                "sharedWithUsers": "Shared with users",
+                "sharedWithGroups": "Also share with…",
+                "editUser": "{{name}} can view & edit",
+                "viewOnlyUser": "{{name}} can view",
+                "removeUser": "{{name}} can't view",
```
ErikMichelson, line 329: Maybe use something like: "Remove {{name}}'s permissions"?

src/components/editor/document-bar/permissions/permission-modal.tsx
```diff
@@ -0,0 +1,162 @@
...
+
+export interface PermissionsModalProps {
+  show: boolean,
+  onChangeShow: (newShow: boolean) => void
+}
+
+export interface Principal {
+  id: string
+  name: string
+  photo: string
```
ErikMichelson, line 19: You could strip down the attributes by extending `UserResponse`.

```suggestion
export interface Principal extends UserResponse {
```

If you apply this change, you need to add the `provider` attribute in the `addUser` method as it is defined in the response as well.

src/components/editor/document-bar/permissions/permission-modal.tsx
```diff
@@ -0,0 +1,162 @@
...
+        photo: response.photo
+      })
+    }).catch(() => setError(true))
+    // set user List
+    permissionsApiResponse.sharedTo.forEach(shareUser => {
+      getUserById(shareUser.username).then(response => {
+        setUserList(list => list.concat([{
+          id: response.id,
+          name: response.name,
+          photo: response.photo,
```
ErikMichelson, line 75: You don't need to specify each attribute of the UserResponse manually if you extend it as described some comments above.

```suggestion
          ...response,
```

ErikMichelson (changes requested):

src/components/editor/document-bar/permissions/permission-modal.tsx
mrdrogdrog: Objection!
A principal ist not an api response. Please don't extend stuff just because they have similar fields. Try to maintain a logic structure.

ErikMichelson: Well, I see your point.

I had the idea of extending UserResponse, because the other attributes except `canEdit` are in fact fetched from the API which returns the response object.

Maybe we should define some common User-attributes which are used by the API response, by the Principal objects, by the Avatar-objects? I don't think it makes much sense to copy the exact same attributes into each new interface that has something to do with users.

mrdrogdrog: That would be a better solution

src/components/editor/document-bar/permissions/permission-list.tsx
DerMolly: done

src/components/editor/document-bar/permissions/permission-modal.tsx
DerMolly: done

src/components/editor/document-bar/permissions/permission-modal.tsx
DerMolly: done

public/locales/en.json
DerMolly: done

src/components/editor/document-bar/permissions/permission-modal.tsx
DerMolly: I'd suggest working that out in separate PR

src/components/editor/document-bar/permissions/permission-modal.tsx
DerMolly: done

src/components/editor/document-bar/permissions/permission-modal.tsx
DerMolly: And with that I won't incorporate these changes as well

src/components/editor/document-bar/permissions/permission-modal.tsx
DerMolly: done

ErikMichelson (approved):

src/components/editor/document-bar/permissions/permission-modal.tsx
DerMolly: done

src/components/editor/document-bar/permissions/permission-modal.tsx
```diff
@@ -0,0 +1,160 @@
...
+        }]))
+      }).catch(() => setError(true))
+    })
+    // set group List
+    permissionsApiResponse.sharedToGroup.forEach(sharedGroup => {
+      if (sharedGroup.id === '1') {
+        setAllUserPermissions(sharedGroup.canEdit ? GroupMode.EDIT : GroupMode.VIEW)
+      } else if (sharedGroup.id === '2') {
+        setAllLoggedInUserPermissions(sharedGroup.canEdit ? GroupMode.EDIT : GroupMode.VIEW)
+      }
```
mrdrogdrog: Magic numbers

mrdrogdrog (approved):

PR #534, opened 2020-09-03
Show warning if note is longer than configured maximum length

ErikMichelson:
### Component/Part
Editor

### Description
This PR adds support for setting a maximal document length and showing warning modals and information if the note is longer than the limit.

### Steps

- [x] added implementation
- [x] added / updated tests
- [ ] extended changelog

### DIFF
--- a/package.json
+++ b/package.json
@@ -147,6 +147,7 @@
         "@types/redux-devtools": "3.0.47",
         "@types/redux-devtools-extension": "2.13.2",
         "cypress": "5.1.0",
+        "cypress-commands": "^1.1.0",
         "eslint-plugin-chai-friendly": "0.6.0",
         "eslint-plugin-cypress": "2.11.1",
         "http-server": "0.12.3",
--- a/public/locales/en.json
+++ b/public/locales/en.json
@@ -277,7 +277,7 @@
             },
             "lines": "{{lines}} Lines",
             "length": "Length {{length}}",
-            "lengthTooltip": "You still may write {{remaining}} characters until you reach the limit for this document."
+            "lengthTooltip": "You may still write {{remaining}} characters until you reach the limit for this note."
         },
         "export": {
             "rawHtml": "Raw HTML",
--- a/public/locales/en.json
+++ b/public/locales/en.json
@@ -196,7 +196,7 @@
             },
             "limitReached": {
                 "title": "Reached the limit",
-                "description": "Sorry, you've reached the maximum length this note can be ({{maxLength}} characters).",
+                "description": "Sorry, this note reached its maximum length. Notes can be up to {{maxLength}} characters long.",
                 "advice": "Please shorten the note."
             },
             "incompatible": {
--- a/public/locales/en.json
+++ b/public/locales/en.json
@@ -277,7 +277,7 @@
             },
             "lines": "{{lines}} Lines",
             "length": "Length {{length}}",
-            "lengthTooltip": "You can write up to 100000 characters in this document."
+            "lengthTooltip": "You still may write {{remaining}} characters until you reach the limit for this document."
         },
         "export": {
             "rawHtml": "Raw HTML",
--- a/public/locales/en.json
+++ b/public/locales/en.json
@@ -277,7 +277,11 @@
             },
             "lines": "{{lines}} Lines",
             "length": "Length {{length}}",
-            "lengthTooltip": "You may still write {{remaining}} characters until you reach the limit for this note."
+            "lengthTooltip": {
+                "maximumReached": "You've reached the maximum length of this note.",
+                "remaining": "You may still write {{remaining}} characters until you reach the limit for this note.",
+                "exceeded": "You've exceeded the maximum length of this note by {{exceeded}} characters. Consider shortening it."
+            }
         },
         "export": {
             "rawHtml": "Raw HTML",
--- a/public/locales/en.json
+++ b/public/locales/en.json
@@ -195,8 +195,8 @@
                 "description": "Sorry, only the owner can edit this note."
             },
             "limitReached": {
-                "title": "Reach the limit",
-                "description": "Sorry, you've reached the maximum length this note can be.",
+                "title": "Reached the limit",
+                "description": "Sorry, you've reached the maximum length this note can be ({{maxLength}} characters).",
                 "advice": "Please shorten the note."
             },
             "incompatible": {
--- a/src/components/editor/editor-modals/max-length-warning-modal.tsx
+++ b/src/components/editor/editor-modals/max-length-warning-modal.tsx
@@ -1,0 +1,27 @@
+import React from 'react'
+import { Button, Modal } from 'react-bootstrap'
+import { Trans, useTranslation } from 'react-i18next'
+import { CommonModal } from '../../common/modals/common-modal'
+
+export interface MaxLengthWarningModalProps {
+  show: boolean
+  onHide: () => void
+  maxLength: number
+}
+
+export const MaxLengthWarningModal: React.FC<MaxLengthWarningModalProps> = ({ show, onHide, maxLength }) => {
+  useTranslation()
+
+  return (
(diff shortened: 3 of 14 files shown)


public/locales/en.json
```diff
@@ -189,8 +189,8 @@
                 "description": "Sorry, only the owner can edit this note."
             },
             "limitReached": {
-                "title": "Reach the limit",
-                "description": "Sorry, you've reached the maximum length this note can be.",
+                "title": "Reached the limit",
+                "description": "Sorry, you've reached the maximum length this note can be ({{maxLength}} characters).",
```
SuperSandro2000, line 193:
```suggestion
                "description": "Sorry, you've reached the maximum length this note can only be ({{maxLength}} characters).",
```

public/locales/en.json
```diff
@@ -189,8 +189,8 @@
...
                 "advice": "Please shorten the note."
```
SuperSandro2000, line 194:
```suggestion
                "advice": "Please shorten the note by removing or splitting content into other notes."
```

public/locales/en.json
```diff
@@ -271,7 +271,7 @@
             },
             "lines": "{{lines}} Lines",
             "length": "Length {{length}}",
-            "lengthTooltip": "You can write up to 100000 characters in this document."
+            "lengthTooltip": "You still may write {{remaining}} characters until you reach the limit for this document."
```
SuperSandro2000, line 274:
```suggestion
            "lengthTooltip": "You may still write {{remaining}} characters until you reach the limit for this note."
```

Not sure on the may still part.

public/locales/en.json
ErikMichelson: I don't think that adding this filler word is really required.

public/locales/en.json
ErikMichelson: Although both variants should be acceptable, the phrase "you may still ..." seems to be better chosen. I'll change that.
The term "document" is taken from the wording in version 1.6. But as we're referring to "notes" everywhere it should probably be used here too.

public/locales/en.json
ErikMichelson: The advice is meant to be a bold, short statement as shown in the screenshot below.
It delivers a short but precise answer what to do to avoid further errors of this kind. I wouldn't say it's necessary to add your suggested explanation.

image

public/locales/en.json
ccoenen: "Sorry, you've reached the maximum length for notes. Notes can be up to {{maxLength}} characters long."

Yes that's even longer, but doesn't sound as terse.

src/components/editor/editor-modals/max-length-warning-modal.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+export const MaxLengthWarningModal: React.FC<MaxLengthWarningModalProps> = ({ show, onHide, maxLength }) => {
+  useTranslation()
+
+  return (
+    <CommonModal show={show} onHide={onHide} titleI18nKey={'editor.error.limitReached.title'} closeButton={true}>
+      <Modal.Body className={'limit-warning'}>
+        <Trans i18nKey={'editor.error.limitReached.description'} values={{ maxLength }} />
+        <br /><br/>
+        <b><Trans i18nKey={'editor.error.limitReached.advice'}/></b>
```
DerMolly, line 20: This should be accomplished by CSS?

package.json
```diff
@@ -143,6 +143,7 @@
         "@types/redux-devtools-extension": "2.13.2",
         "@types/testing-library__cypress": "5.0.6",
         "cypress": "5.0.0",
+        "cypress-commands": "^1.1.0",
```
DerMolly, line 146: As far as I can see your not defining any custom commands so this seems to be unnecessary (at least for the moment)

DerMolly (changes requested):

src/components/editor/editor-modals/max-length-warning-modal.tsx
mrdrogdrog: Use <strong>

DerMolly: I also meant the `<br/>`s

mrdrogdrog (changes requested):
1. I don't know if the start of the document is a good place for the warning, because you can see it only when you're at the top.

2. The modal is a bit annoying when you try to remove the redundant characters, because you have to dismiss it for every one. It would be better if you remember that it was already shown until you reduce the character count and then show it again as soon as you exceed again.

package.json
ErikMichelson: Sorry, that's wrong. I'm using the `.attribute()` selector in the cypress test to test the title attribute of the status-bar length. That selector isn't included in the core-package of cypress but is in the cypress-commands package. See [the docs](https://github.com/Lakitna/cypress-commands/blob/develop/docs/attribute.md#attribute) for details about it.

ErikMichelson:
@mrdrogdrog 
> 1. I don't know if the start of the document is a good place for the warning, because you can see it only when you're at the top.

Should I move the warning to the bottom or add another warning at the bottom?

> 2. The modal is a bit annoying when you try to remove the redundant characters, because you have to dismiss it for every one. It would be better if you remember that it was already shown until you reduce the character count and then show it again as soon as you exceed again.

That's a good point. I'll change that behaviour.

src/components/editor/editor-modals/max-length-warning-modal.tsx
mrdrogdrog: Well yes. br should be used for actual line breaks. For margin please use the margin classes.

ErikMichelson: Done in aef6aac

DerMolly (approved):

SISheogorath:
Question: Should it be a popup warning or a banner warning? I could also imagine a banner at the top that warns about such things which would be a bit less jumping into people's face than a popup and contribute to a distraction free environment. This would also allow to start a yellow warning banner with 95k or 95% of a document is filled with text, informing people that they might have to take action soon.

src/components/editor/editor-pane/editor-pane.tsx
```diff
@@ -99,15 +105,24 @@ export const EditorPane: React.FC<EditorPaneProps & ScrollProps> = ({ onContentC
   }, [editor, scrollState])
 
   const onBeforeChange = useCallback((editor: Editor, data: EditorChange, value: string) => {
+    if (value.length > maxLength && !maxLengthWarningAlreadyShown.current) {
```
SISheogorath, line 108: Can this trigger when one doesn't have edit rights?

mrdrogdrog:
First there is a color indicator in the editor, that turns yellow and then red. But if you really exceed the character count, then the modal will pop up. IMHO this is okay, because we need an error message that interrupts the user and get his attention.

src/components/editor/editor-pane/editor-pane.tsx
mrdrogdrog: Yes. Because right now we have no information if a user is allowed to edit a doc or not.

mrdrogdrog:
image

Well.. technically yes.. but not helpful :D

ErikMichelson:
@mrdrogdrog Check the tooltip-title again. I've now added additional messages on reaching and exceeding the limit.

src/components/editor/editor-pane/editor-pane.tsx
ErikMichelson: I'm unsure if the `onBeforeChange` event really will fire when the user doesn't change anything. I would guess it won't be triggered by "external" changes of the value.

mrdrogdrog (approved):

src/components/editor/editor-pane/editor-pane.tsx
mrdrogdrog: onBeforeChange will only be triggered if you type something.

mrdrogdrog: So correction: No. This won't trigger for read only access.

PR #551, opened 2020-09-08
Extract URLs in extra file and change riot link to element-link

ErikMichelson:
### Component/Part
Intro page -> Footer links, Error page link

### Description
This PR extracts static URLs throughout the application into an extra file for easier changes in the future. It also changes the former riot.im link to the new app.element.io link for our chat room.

### Steps

- [x] added implementation
- [x] added / updated tests
- [ ] added / updated documentation _(nothing to document here)_
- [ ] extended changelog _(probably not worth a changelog entry?)_

### Related Issue(s)
codimd/server#495

### DIFF
--- a/cypress/integration/link.spec.ts
+++ b/cypress/integration/link.spec.ts
@@ -148,7 +148,7 @@ describe('Links Intro', () => {
     })
 
     it('Matrix', () => {
-      cy.get('a[href="https://riot.im/app/#/room/#hedgedoc:matrix.org"]')
+      cy.get('a[href="https://app.element.io/#/room/#hedgedoc:matrix.org"]')
         .checkExternalLink('https://app.element.io/#/room/#hedgedoc:matrix.org')
     })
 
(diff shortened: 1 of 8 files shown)


DerMolly:
@SISheogorath Could we just use https://hedgedoc.com/matrix and let the exact link the be decided there?

ErikMichelson:
I would rather use matrix.hedgedoc.org to continue the scheme (translation., community., ...)

DerMolly:
Also it could be a good idea to extract all static links in a separate json file and import what we need from there. This way changing something along those lines is easiest in the future.

ErikMichelson:
@DerMolly Would that be another PR or should it be introduced with this PR?

DerMolly:
If we want to do that, we could do that here imho.

DerMolly (approved):

mrdrogdrog (approved):

PR #563, opened 2020-09-12
Add read-only view (/s/note)

### LINKED ISSUE #367: View-Only-Mode
Codimd had a "view-only" mode (`/s/`). We should reimplement this.

ErikMichelson:
### Component/Part
 - The renderer (restructured)
 - The render-pane (restructured)
 - The router (added route)
 - The read-only-view (added)

### Description
This PR adds the read-only-view accessible through `/s/note-id`.

### Steps

- [x] added implementation
- [ ] added / updated tests
- [x] extended changelog

### Related Issue(s)
Closes #367


src/components/editor/app-bar/app-bar.tsx
```diff
@@ -25,15 +29,19 @@ export const AppBar: React.FC = () => {
...
         <DarkModeButton/>
         <Link to={`/p/${id}`} target='_blank'>
           <Button title={t('editor.documentBar.slideMode')} className="ml-2 text-secondary" size="sm" variant="outline-light">
             <ForkAwesomeIcon icon="television"/>
           </Button>
         </Link>
-        <HelpButton/>
+        <ShowIf condition={showEditorButtons}>
+          <HelpButton/>
+        </ShowIf>
```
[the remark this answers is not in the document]
mrdrogdrog: Well the help stuff is mostly for edit mode

src/components/editor/document-bar/document-info/document-info-line.tsx
```diff
@@ -4,12 +4,13 @@ import { IconName } from '../../../common/fork-awesome/types'
 
 export interface DocumentInfoLineProps {
   icon: IconName
+  isInline?: boolean
```
mrdrogdrog, line 7: IMHO "isInline" is a weird name for a variable that sets the icon size. How about just providing the icon size as optional argument? That would take a huge chunk of complexity out of this simple component.

src/components/editor/document-bar/document-info/document-info-time-line.tsx
```diff
@@ -19,16 +20,16 @@ export enum DocumentInfoLineWithTimeMode {
   EDITED
 }
 
-export const DocumentInfoTimeLine: React.FC<DocumentInfoLineWithTimeProps> = ({ time, mode, userName, profileImageSrc }) => {
+export const DocumentInfoTimeLine: React.FC<DocumentInfoLineWithTimeProps> = ({ time, mode, userName, profileImageSrc, isInline }) => {
   useTranslation()
 
   const i18nKey = mode === DocumentInfoLineWithTimeMode.CREATED ? 'editor.modal.documentInfo.created' : 'editor.modal.documentInfo.edited'
   const icon: IconName = mode === DocumentInfoLineWithTimeMode.CREATED ? 'plus' : 'pencil'
```
mrdrogdrog, line 27: I'm not happy anymore with the DocumentInfoLineWithTimeMode, because in react you should create generic components and then more specific ones. This means that we would need a "DocumentInfoTimeFromNowLine" (or something else), and then two specific components for "created" and "edited". Yes, I know that I created this component.. but I think it would be more correct. If you agree I would change this on my own.

src/components/editor/document-bar/document-info/document-info-time-line.tsx
```diff
@@ -19,16 +20,16 @@ export enum DocumentInfoLineWithTimeMode {
...
 
   return (
-    <DocumentInfoLine icon={icon}>
+    <DocumentInfoLine icon={icon} isInline={isInline}>
       <Trans i18nKey={i18nKey} >
-        <UserAvatar photo={profileImageSrc} additionalClasses={'document-info-avatar font-style-normal bold font-weight-bold'} name={userName}/>
+        <UserAvatar photo={profileImageSrc} additionalClasses={`font-style-normal bold font-weight-bold ${!isInline ? 'document-info-avatar' : ''}`} name={userName}/>
```
mrdrogdrog, line 32: Here is this "isInline" again. The variable name doesn't say anything about what it does. How about  a "small/big user avatar" prop for the useravatar component instead?

src/components/editor/document-renderer-pane/scrolling-document-render-pane.tsx
```diff
@@ -0,0 +1,103 @@
+import React, { useCallback, useEffect, useRef, useState } from 'react'
+import { LineMarkerPosition } from '../../markdown-renderer/types'
+import { ScrollProps, ScrollState } from '../scroll/scroll-props'
+import { findLineMarks } from '../scroll/utils'
+import { DocumentRenderPane, DocumentRenderPaneProps } from './document-render-pane'
+
+export const ScrollingDocumentRenderPane: React.FC<DocumentRenderPaneProps & ScrollProps> = ({
```
mrdrogdrog, line 7: This component does so much more then just scrolling.. But right know I don't have an idea for a better structure.

src/components/markdown-renderer/full-markdown-renderer.tsx
```diff
@@ -1,227 +1,61 @@
...
 export interface FullMarkdownRendererProps {
-  onFirstHeadingChange?: (firstHeading: string | undefined) => void
   onLineMarkerPositionChanged?: (lineMarkerPosition: LineMarkerPosition[]) => void
+  onFirstHeadingChange?: (firstHeading: string | undefined) => void
   onMetaDataChange?: (yamlMetaData: YAMLMetaData | undefined) => void
   onTaskCheckedChange: (lineInMarkdown: number, checked: boolean) => void
   onTocChange?: (ast: TocAst) => void
 }
 
 export const FullMarkdownRenderer: React.FC<FullMarkdownRendererProps & AdditionalMarkdownRendererProps> = ({
```
mrdrogdrog, line 19: what is the difference between the full markdown renderer and the enhanced one?..

src/components/editor/app-bar/app-bar.tsx
```diff
@@ -16,7 +16,11 @@ import { EditorViewMode } from './editor-view-mode'
 import { HelpButton } from './help-button/help-button'
 import { NavbarBranding } from './navbar-branding'
 
-export const AppBar: React.FC = () => {
+export interface AppBarProps {
+  showEditorButtons: boolean
+}
+
+export const AppBar: React.FC<AppBarProps> = ({ showEditorButtons }) => {
```
mrdrogdrog, line 23: I don't know if it's a good idea to use one bar for two views. Especially if one of them is having less buttons. Maybe we should create a second component for that?

mrdrogdrog (changes requested):

src/components/pad-view-only/document-infobar.tsx
```diff
@@ -0,0 +1,62 @@
...
+  changedAuthor: string
+  changedTime: Date
+  createdAuthor: string
+  createdTime: Date
+  editable: boolean
+  noteId: string
+  viewCount: number
+}
+
+export const DocumentInfobar: React.FC<DocumentInfobarProps> = ({
```
mrdrogdrog, line 22: tbh I don't like this info lines. They are so codimd 1.6..  that's why we replaced them in the editor view with the document info modal. I would use that modal here, too.  (my opinion)

src/components/editor/app-bar/app-bar.tsx
DerMolly: As it is the same bar but for another purpose I quite like Eriks solution here… Two bars would make it hard to maintain. I think it's okay like this. One could argue a comment might improve the understanding of the prop, but I think it's fine like this.

mrdrogdrog: Okay but then I would rather use an enum than a boolean.

DerMolly: You mean something like AppBarMode?

mrdrogdrog: Yes. I mean it's @ErikMichelson s decision.. But imho an enum would be better because then the purpose of this argument is clearer.

DerMolly: Hm, an enum implies more than to modes and that could be the point where different bars are a good idea… 🤔

src/components/editor/app-bar/app-bar.tsx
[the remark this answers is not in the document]
ErikMichelson: That's the point. If you see my commit history, I firstly didn't make the help dialog optional but except of the "links" page, everything else is not appliable to the read-only view. I therefore removed it.

src/api/notes/index.ts
```diff
@@ -2,7 +2,6 @@ import { defaultFetchConfig, expectResponseCode, getApiUrl } from '../utils'
 
 interface LastChange {
```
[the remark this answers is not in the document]
ErikMichelson: I'm still quite unhappy about the whole Notes-interface. In my opinion it's too directly copied from 1.6 and doesn't respect newer things like the permissions of a note etc.

I just didn't want to change that in this PR, because I guess it'll be a more relevant problem/topic of the backend integration.

src/components/editor/document-bar/document-info/document-info-line.tsx
ErikMichelson: Good point. I initially also changed the font-size with this setting but as setting the font-size to inherit is enough to make it work in all places, I'm going to change that as you mentioned it here.

src/components/editor/document-bar/document-info/document-info-time-line.tsx
ErikMichelson: The difference is just the translation and the icon. I can understand that it doesn't follow the strict react-recommendations/rules but I fear that adding two more components that just contain their specific values are more complex to read/understand?

Anyway, if you want to change it, I'm not going to stop you.

src/components/editor/document-bar/document-info/document-info-time-line.tsx
ErikMichelson: Ha, I knew there was something else with the isInline, but thought I removed that also.

I'll change it.

src/components/editor/app-bar/app-bar.tsx
ErikMichelson: I don't know which features else could be optional in the bar but not-editor specific. But on the other side I can't predict the future. And changing that into an enum is a rather simple task. I'll do that.

src/components/pad-view-only/document-infobar.tsx
ErikMichelson: You mean including the document-info button into the app-bar? That's a no-go in my opinion as it violates the document/app diffentiation.

ErikMichelson: Of course, I only can talk for myself. But when someone sends me a link to a read-only-version of a HedgeDoc/CodiMD document, I like to see directly how up-to-date it is and who created or edited it last. Clicking another button just makes it more inconvenient to get that information.

In the editor, I can fully live with the button-solution. It even looks better because of not cluttering the UI with information everywhere. But the read-only-view is another thing. I almost would go as far as comparing it with a blog post or medium article - in both of those two you get presented the author and creation-time without further action.

To conclude: I wouldn't change that.

src/components/markdown-renderer/full-markdown-renderer.tsx
ErikMichelson: The enhanced renderer extends the basic renderer by enabling diagrams, one-click-components and all the other fancy stuff of HedgeDoc.

The full renderer extends the enhanced renderer by handling the line-number-markers and adding the line-marker replacer.

src/components/editor/app-bar/app-bar.tsx
ErikMichelson: Done in a10ebd4

src/components/editor/document-bar/document-info/document-info-time-line.tsx
ErikMichelson: Done in 54a91cb

src/components/editor/document-bar/document-info/document-info-line.tsx
ErikMichelson: Done in 4695bae

src/components/editor/app-bar/app-bar.tsx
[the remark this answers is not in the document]
DerMolly: Okay fair point

DerMolly (approved):

mrdrogdrog:
rebase plz

ErikMichelson:
@mrdrogdrog Another rebase is done

mrdrogdrog (changes requested):

mrdrogdrog:
I know you'll probably hate me, but maybe we should merge https://github.com/codimd/react-client/pull/626 first? It moves A LOT of logic out of the components and could make this PR a log smaller.

ErikMichelson:
Okay let's do that with #626 first.

src/components/markdown-renderer/enhanced-markdown-renderer.tsx
```diff
@@ -0,0 +1,245 @@
...
+          const meta: RawYAMLMetadata = yaml.safeLoad(rawMeta) as RawYAMLMetadata
+          setYamlError(false)
+          rawMetaRef.current = meta
+        } catch (e) {
+          console.error(e)
+          setYamlError(true)
+          rawMetaRef.current = ({} as RawYAMLMetadata)
+        }
+      })
+    }
```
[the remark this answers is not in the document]
ErikMichelson: Oh, rebasing is always funny. 😅

src/components/editor/document-renderer-pane/scrolling-document-render-pane.tsx
DerMolly: what does it do besides scrolling? It seems to me that's more or less it purpose…

DerMolly (approved):

src/components/editor/document-bar/document-info/document-info-time-line.tsx
```diff
@@ -19,16 +19,16 @@ export enum DocumentInfoLineWithTimeMode {
...
 
   const i18nKey = mode === DocumentInfoLineWithTimeMode.CREATED ? 'editor.modal.documentInfo.created' : 'editor.modal.documentInfo.edited'
   const icon: IconName = mode === DocumentInfoLineWithTimeMode.CREATED ? 'plus' : 'pencil'
 
   return (
-    <DocumentInfoLine icon={icon}>
+    <DocumentInfoLine icon={icon} size={size}>
       <Trans i18nKey={i18nKey} >
-        <UserAvatar photo={profileImageSrc} additionalClasses={'document-info-avatar font-style-normal bold font-weight-bold'} name={userName}/>
+        <UserAvatar photo={profileImageSrc} additionalClasses={`font-style-normal bold font-weight-bold ${size ? 'document-info-avatar' : ''}`} name={userName} size={size ? 'lg' : undefined}/>
```
mrdrogdrog, line 31: Is the condition for `document-info-avatar` still needed? You deleted the scss file that contained this rule.

mrdrogdrog (changes requested):

src/components/editor/document-bar/document-info/document-info-time-line.tsx
ErikMichelson: Seems like a mistake, I check it

ErikMichelson: Removed

mrdrogdrog (approved):

PR #568, opened 2020-09-15
Add caching of user-data

ErikMichelson:
### Component/Part
API requests for user profile data

### Description
This PR adds caching-logic for user profile data to not send the same request too often to the backend.

### Steps

- [x] added implementation
- [ ] added / updated tests
- [ ] extended changelog

### Related Issue(s)
none


src/api/users/index.ts
```diff
@@ -1,10 +1,22 @@
 import { defaultFetchConfig, expectResponseCode, getApiUrl } from '../utils'
-import { UserResponse } from './types'
+import { UserResponse, UserResponseCacheEntry } from './types'
+
+const CACHE_TIME_SECONDS = 600
```
DerMolly, line 4: This could be moved into a cache-time.json file. Especially if we have more caches in the future and they may have different times?

src/api/users/types.d.ts
```diff
@@ -6,3 +6,8 @@ export interface UserResponse {
   photo: string
   provider: LoginProvider
 }
+
+export interface UserResponseCacheEntry {
```
DerMolly, line 10: In the same vein this could be a a generic CacheEntry<T>…

DerMolly (changes requested):

src/api/users/index.ts
ErikMichelson: But every cache will be bound to an API request method, doesn't it?
In that case, I would argue that having the cache time definition in the same place as the caching logic is more intuitive.

src/api/users/types.d.ts
ErikMichelson: That's true. And I even can move the revision cache into the API-logic and let that use a common cache entry object.

ErikMichelson: Done

src/api/users/index.ts
DerMolly: I like the idea of changing the cache times in a single file, but I guess it doesn't happen to often.

ErikMichelson: That's the point. And even if you change the times, that will be opening maybe 3-4 files.
The single static links collection is another topic because they're used over the whole project.

DerMolly (approved):

mrdrogdrog (changes requested):
I see many potential duplicate code here. How about creating a class "ResponseCache" that holds all the logic and the map?

mrdrogdrog (changes requested):
Please solve the linting errors before I do review.

mrdrogdrog:
Sry. Github showed an old version -_-. Ignore the last comment.

mrdrogdrog (changes requested):
Unit (not integration) tests for the cache would be nice.

src/components/common/cache/Cache.test.ts
```diff
@@ -0,0 +1,72 @@
...
+
+  it('entry limit is respected', () => {
+    const limitedCache = new Cache<string, number>(1000, 2)
+    limitedCache.put('first', 1)
+    expect(limitedCache.has('first')).toBe(true)
+    limitedCache.put('second', 2)
+    expect(limitedCache.has('second')).toBe(true)
+    limitedCache.put('third', 3)
+    expect(limitedCache.has('third')).toBe(true)
+    expect(limitedCache.has('first')).toBe(false)
```
DerMolly, line 70:
```suggestion
   expect(limitedCache.has('second')).toBe(true)
    expect(limitedCache.has('first')).toBe(false)
```

DerMolly (changes requested):

src/components/common/cache/Cache.test.ts
ErikMichelson: I made the checks for this test more explicit in dc5fc81.

mrdrogdrog (approved):

DerMolly (approved):

PR #594, opened 2020-09-20
fixed highlight fences showLines and wrapLines

### LINKED ISSUE #592: Flags have no effect in code fences
**Description**
The code fences are always showing line numbers. They should do that only if the language ends with an `=`.

**To Reproduce**
1. open editor
2. create an code fence with any language
3. don't use `=`

**Expected behavior**
You can toggle between line numbers on and off.

DerMolly:
### Component/Part
renderer

### Description
This PR fixes the highlight fences showLines and wrapLines.

Please note: cypress does currently not support copy to clipboard in firefox so this part of the test is commented out (for the time being)

### Steps

- [ ] added implementation
- [x] added / updated tests
- [ ] extended changelog

### Related Issue(s)
fixes #592

### DIFF
--- a/package.json
+++ b/package.json
@@ -155,7 +155,6 @@
         "@testing-library/cypress": "7.0.0",
         "@types/redux-devtools": "3.0.47",
         "@types/redux-devtools-extension": "2.13.2",
-        "clipboardy": "^2.3.0",
         "cross-env": "7.0.2",
         "cypress": "5.2.0",
         "cypress-commands": "1.1.0",
--- a/package.json
+++ b/package.json
@@ -155,6 +155,7 @@
         "@testing-library/cypress": "7.0.0",
         "@types/redux-devtools": "3.0.47",
         "@types/redux-devtools-extension": "2.13.2",
+        "clipboardy": "^2.3.0",
         "cross-env": "7.0.2",
         "cypress": "5.2.0",
         "cypress-commands": "1.1.0",
--- a/cypress/integration/code.spec.ts
+++ b/cypress/integration/code.spec.ts
@@ -81,7 +81,7 @@ describe('Code', () => {
     cy.get('.markdown-body > pre > div > button > i')
       .should('have.class', 'fa-files-o')
       .click()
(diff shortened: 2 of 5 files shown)


ErikMichelson (approved):

package.json
```diff
@@ -155,6 +155,7 @@
         "@testing-library/cypress": "7.0.0",
         "@types/redux-devtools": "3.0.47",
         "@types/redux-devtools-extension": "2.13.2",
+        "clipboardy": "^2.3.0",
```
mrdrogdrog, line 158: why do we need this?

mrdrogdrog (changes requested):

package.json
DerMolly: this is used to get the clipboard from node to check if the copy worked. The cypress test is currently deactivated as firefox does not copy in cypress currently, but chrome does work. I committed the test commented out for the time when firefox can copy things in cypress. It's used in the plugin.js file in cypress

mrdrogdrog: So you created a test, that isn't activated because it doesn't work?
IMHO it would be better to create an issue (and add a todo to the test?), that says "create the test as soon as cypress+firefox is ready for clipboard"?
Because right now we have an unused dependency.

DerMolly: I created a test with stuff I found in issues about the problem, so I'm very reluctant to remove that working test. If you don't like the dependency we could remove it and add it to the comment in the test.

mrdrogdrog: Well but the test doesn't work because the code is deactivated, right?

mrdrogdrog: But if that test would work it wouldn't be deactivated , right? 😅

DerMolly: As stated above one of the tests does not work with firefox, because privacy and cypress did not (yet) build a workaround for that.

DerMolly: Other than that the test is working flawless.

DerMolly: I did what I suggested

mrdrogdrog (approved):

PR #616, opened 2020-09-30
Extract clean title from first heading

### LINKED ISSUE #440: Exclude tags (especially images) from document name
I recently tried adding a small logo to one of my documents with the  syntax and while that works flawlessly I also noticed that this would display the un-interpreted line as the "document name" in the personal history. That basically makes the title unreadable. Other common markdown tags such as italic, bold etc are already ignored as far as I can tell.

So my feature request would be to remove interpretable tags from the header alltogether.

DerMolly:
### Component/Part
editor title extraction

### Description
This PR improves the title extraction from the first heading.

### Steps

- [x] added implementation
- [x] added / updated tests
- [ ] extended changelog

### Related Issue(s)
fixes #440


src/components/editor/utils.ts
```diff
@@ -1 +1,15 @@
...
+
+export const removeMarkdownSyntax = (input: string): string => {
+  return input
+    .replace(/\*(.*?)\*/gi, '$1') // bold & italic
+    .replace(/\+\+(.*?)\+\+/gi, '$1') // underline
+    .replace(/~(.*?)~/gi, '$1') // subscript & strikethrough
+    .replace(/\^(.*?)\^/gi, '$1') // superscript
+    .replace(/`(.*?)`/gi, '$1') // code
+    .replace(/!?\[(.*?)]\(.*?\)/gi, '$1') // link & image
+}
```
mrdrogdrog, line 15: You're regex are not working 100% perfect.
e.g. if you
- input `1*2*3 4*5**` then the result should be `123 4*5**` but it is `123 45*`
- input `**1 2*` then the result should be `*1 2` but it is `1 2*`

As much as I appreciate that you try to keep this simple and independent from other components, I think a better solution would  to use the DOM results, because then we have the output from the actual markdown renderer.

mrdrogdrog: (also you forgot `_asd_`)

mrdrogdrog (changes requested):

src/components/editor/utils.ts
DerMolly: fixed by using the render output

mrdrogdrog (approved):

ErikMichelson (approved):

Issue #430, opened 2020-08-17
Add an own emoji-picker as emoji-mart replacement


ErikMichelson:
**What is your question?**
Currently the react-client uses [emoji-mart](https://github.com/missive/emoji-mart). Because that has neither the current emoji definitions nor is maintained actively, @mrdrogdrog, @DerMolly and me looked for alternatives.
One idea was to use the emoji picker from element-web but as it is heavy integrated into the matrix-react-sdk, that's also no solution.
Even if it's quite some work, we thought that building an own emoji-picker (in a separate repository) is probably the best solution.
As data source we could use this one: https://github.com/milesj/emojibase/

Other opinions on that?

mrdrogdrog:
Another option was to fork and maintain emoji-mart. But that code is awfull and we would have to rewrite half of the code we need.  Also we don't want to maintain the stuff that we don't need.

mrdrogdrog:
Well. We could use the matrix react sdk if we turn HedgeDoc into an collaborative markdown matrix client 😏

mrdrogdrog:
https://github.com/nolanlawson/emoji-picker-element

ccoenen:
I have also raised the question in riot, if this is a must-have / must replace for 2.0. Especially if it becomes neccessary to develop something entirely new, I'd suggest putting this in a release after 2.0.

DerMolly:
That entirely depends on the schedule for 2.0 if we target a release in December or so, yes. If the release is more in the vain of April 2021 we‘ll have time to do that for 2.0.

mrdrogdrog:
TBH: we're running out of issues that are 2.0 relevant but can be solved until the backend is ready. So there is a good chance that this issue will be part of 2.0.

mrdrogdrog:
> https://github.com/nolanlawson/emoji-picker-element

This picker doesn't support custom css classes for emojis, which is necessary for fork-awesome. Who wants to create an issue or talk to the dev? @ErikMichelson  ?

PR #620, opened 2020-10-01
Replace emoji-mart with emoji-picker-element

ErikMichelson:
### Component/Part
Editor -> Toolbar -> Emoji picker

### Description
This PR replaces emoji-mart with emoji-picker-element. Please note, that this doesn't already include the ForkAwesome icons correctly as we wait for response from the emoji-picker-element developers.

### Steps

- [x] added implementation
- [x] added / updated tests
- [x] extended changelog

### Related Issue(s)
#430 (not closing)


src/components/editor/editor-pane/tool-bar/emoji-picker/emoji-picker.tsx
```diff
@@ -1,47 +1,81 @@
...
+  useEffect(() => {
+    if (!pickerContainerRef.current || firstOpened.current) {
+      return
+    }
+    const picker = new Picker({
+      customEmoji: customEmojis,
+      dataSource: '/static/js/emoji-data.json'
+    })
+    const container = pickerContainerRef.current
+    picker.addEventListener('emoji-click', emojiClickListener)
```
DerMolly, line 52: remove this listener onUnload?

DerMolly (changes requested):

DerMolly:
Also this PR should probably remove `externalTypes/emoji-mart`

ErikMichelson:
It does remove the external types: https://github.com/codimd/react-client/pull/620/commits/b109987f4a536f27e7f638c160772b4a72179ac9#diff-134d226e34bcf57915f0e22064d5d18c

src/components/editor/editor-pane/tool-bar/emoji-picker/emoji-picker.tsx
ErikMichelson: This doesn't work unfortunately. Emojis get unselectable from the picker when applying the following change:

```diff
index 3b98108..d42c743 100644
--- a/src/components/editor/editor-pane/tool-bar/emoji-picker/emoji-picker.tsx
+++ b/src/components/editor/editor-pane/tool-bar/emoji-picker/emoji-picker.tsx
@@ -55,6 +55,9 @@ export const EmojiPicker: React.FC<EmojiPickerProps> = ({ show, onEmojiSelected,
     }
     container.appendChild(picker)
     firstOpened.current = true
+    return () => {
+      picker.removeEventListener('emoji-click', emojiClickListener)
+    }
   }, [pickerContainerRef, emojiClickListener, darkModeEnabled, twemojiStyle])

   useEffect(() => {
```

ErikMichelson: Apparently the useEffect expires already when the picker is opened.

DerMolly: okay then I'll close this

DerMolly (approved):

mrdrogdrog (changes requested):
Something seems to be wrong.

image

mrdrogdrog:
Interessting. It was my local setup that caused the bug. Partly.

I had a local installation of Twemoji to render emojis in my WM with the Name "Twemoji Mozilla". Chrome also uses this file, but can only use the black/white fallback. I really don't know why, because if you load a colored font via css then everything is okay.
The emoji picker has an built-in check to determine if the browser supports colored emojis. It renders an emoji onto two canvas in white and black. If the two renderings are the same (because the browser ignores the font-color and uses the color layers from the font), then the picker assumes, that the browser supports colored fonts.

Now comes the twisted part:
Our CSS imports the font file under the name "twemoji". This is my I see colored emojis. But the picker ignores the custom font-family and [uses a constant font-family for the test](https://github.com/nolanlawson/emoji-picker-element/blob/master/src/picker/utils/testColorEmojiSupported.js). And in this font-family const was?.. right. "Twemoji Mozilla".
So my browser used the local font to do the test, but uses our provided font to do the rendering.

mrdrogdrog:
https://github.com/codimd/react-client/pull/664 fixes this issue.

mrdrogdrog (approved):
I tested it in firefox and it looks good. It's your choice if this PR gets merged before or after #664

PR #626, opened 2020-10-01
markdown-it-configurator 

DerMolly:
### Component/Part
markdown renderer

### Description
This PR moves a lot of code from basic- and full-markdown-renderer to the markdown-it-configurator making them much smaller.
This should help with HTML Exports

### Steps

- [x] added implementation
- [ ] added / updated tests
- [ ] extended changelog


src/components/markdown-renderer/markdown-it-configurator.ts
```diff
@@ -0,0 +1,181 @@
...
+  })
+  md.use(lineNumberMarker(), {
+    postLineMarkers: (lineMarkers) => {
+      onLineMarkers(lineMarkers)
+    }
+  })
+  return md
+}
+
+export const configureLinkfyExtra = (md: MarkdownIt): MarkdownIt => {
```
ErikMichelson, line 174: Maybe rename this to something like postMarkdownConfigurator or similar? It does hold more than only the linkify-extra module.

mrdrogdrog (approved):

src/components/editor/app-bar/help-button/cheatsheet.tsx
```diff
@@ -31,11 +31,10 @@ export const Cheatsheet: React.FC = () => {
...
 
-  const markdownItPlugins = useCallback((md: MarkdownIt) => {
-    validAlertLevels.forEach(level => {
-      md.use(markdownItContainer, level, { render: createRenderContainer(level) })
-    })
-  }, [])
+  const md = basicMarkdownRender()
+  validAlertLevels.forEach(level => {
+    md.use(markdownItContainer, level, { render: createRenderContainer(level) })
+  })
```
mrdrogdrog, line 37: useMemo

src/components/markdown-renderer/basic-markdown-renderer.tsx
```diff
@@ -36,53 +25,25 @@ export const BasicMarkdownRenderer: React.FC<BasicMarkdownRendererProps & Additi
...
   const lastUsedLineId = useRef<number>(0)
 
   const markdownReactDom: ReactElement[] = useMemo(() => {
     if (onBeforeRendering) {
       onBeforeRendering()
     }
+    let md: MarkdownIt = basicMarkdownRender()
+    if (markdownIt) {
+      md = markdownIt
+    }
```
mrdrogdrog, line 44: Why are you creating a markdown renderer in every case? IMHO the markdown it prop shouldn't be optional.

mrdrogdrog (changes requested):

src/components/markdown-renderer/basic-markdown-renderer.tsx
```diff
@@ -26,7 +15,7 @@ import { calculateNewLineNumberMapping } from './utils/line-number-mapping'
 
 export interface BasicMarkdownRendererProps {
   componentReplacers?: ComponentReplacer[],
-  onConfigureMarkdownIt?: (md: MarkdownIt) => void,
+  markdownIt?: MarkdownIt,
```
mrdrogdrog, line 18: IMHO the markdown it prop shouldn't be optional.

mrdrogdrog (changes requested):

src/components/markdown-renderer/markdown-it-configurator.ts
mrdrogdrog: How about a class structure for this? MarkdownConfigurator as an abstract class. BasicMarkdownConfigurator extends this. FullMarkdownConfigurator extends this.

mrdrogdrog: This could make this hole file and the flow a bit cleaner

src/components/markdown-renderer/basic-markdown-renderer.tsx
DerMolly: But it needs to be. When I want to use a BasicMarkdownRender by itself it needs to configure its own markdownit

mrdrogdrog: The react way to do this would be to create a "MarkdownRenderer" component that takes a markdown it prop and a "BasicMarkdownRenderer" that configures a plain markdown it instance and passes it into a "MarkdownRenderer".

mrdrogdrog: Also: I don't see that BasicMarkdownRenderer is used without the prop.

DerMolly: it should be in the help dialog.

mrdrogdrog: Instead of making the prop optional I would introduce a "plain markdown renderer" that uses the basicmarkdownrenderer

ErikMichelson: On the one hand, I understand that another renderer would be better. But remember with the view-only mode I'm gonna separate the full-renderer once again into full- and scrollable-full-renderer. Anytime soon, we'll have more markdown-renderer-components than other components...

mrdrogdrog: So what? That is the purpose of a component system. To split up the tasks. You should use composition.

mrdrogdrog: And "there will be more renderer than other components" is a bit exaggerated 🤨. There will be the basic and this is used by plain, full (which is used by full-scrollable)

ErikMichelson: Composition is one thing and I'm okay with it, if you do it the way you just detailed.
I just want to remind that during software development one should be careful about the maintainability. For new contributors, especially ones without greater React knowledge, it might seem too complex to work on these parts of the software.

mrdrogdrog: IMHO it's more confusing that the basic renderer takes an markdown it instance or not

src/components/markdown-renderer/hooks/use-extract-first-headline.ts
```diff
@@ -0,0 +1,24 @@
...
+  }, [])
+
+  useEffect(() => {
+    if (onFirstHeadingChange && documentElement.current) {
+      const firstHeading = documentElement.current.getElementsByTagName('h1').item(0)
+      if (firstHeading) {
+        onFirstHeadingChange(extractInnerText(firstHeading))
+      }
+    }
+  }, [documentElement, extractInnerText, onFirstHeadingChange])
```
DerMolly, line 23: this needs to be given the content and the documentElement as seen here
https://github.com/codimd/react-client/blob/b094fe1de66cb0124842c667a5cb9bc27b95e844/src/components/markdown-renderer/full-markdown-renderer.tsx#L133-L140
This code isn't called when the markdown content changes and therefore only extracts the first title when the pad get's opened, hence the cypress errors

mrdrogdrog: fixed

DerMolly:
I think this can be merged now

ErikMichelson (changes requested):

ErikMichelson (approved):

mrdrogdrog (approved):

Issue #622, opened 2020-10-01
Readd Import Menu


DerMolly:
We need to add functionality to the import menu:

- [ ] Dropbox
- [ ] Gist
- [ ] Gitlab
- [ ] Clipboard

DerMolly:
Import via Clipboard seems to be a bit complicated

https://developer.mozilla.org/en-US/docs/Web/API/Clipboard#Clipboard_availability

1.6 does is this way

![image](https://user-images.githubusercontent.com/3978662/95011628-44f57300-0632-11eb-87a0-12724e6b3e6a.png)

maybe we want to drop this?

mrdrogdrog:
Caniuse also says "no ie and Firefox".
Tbh "paste your markdown content here " is redundant. And how does the grab from a webpage works?

https://caniuse.com/?search=Clipboard%20api

ErikMichelson:
The webpage grabbing will probably result in CORS errors.

mrdrogdrog:
Except if the server grabs the stuff.

mrdrogdrog:
I vote for drop.

DerMolly:
I think the grab from webpage could just be meant to be HTML

PR #645, opened 2020-10-05
Added markdown-file import

ErikMichelson:
### Component/Part
Editor -> Document-Bar -> Import -> Markdown file

### Description
This PR adds support for importing markdown-files by appending their text content to the current note content.

### Steps

- [x] added implementation
- [x] added / updated tests
- [ ] extended changelog

### Related Issue(s)
#622


DerMolly:
As we use for the export of markdown files just the text `Markdown` maybe this would suffice for the import, too?

ErikMichelson:
I explicitly decided for "Markdown file".
In the export context it's quite clear what the Markdown button does especially because of the heading _download_ above it.
As HedgeDoc primarily is a markdown note editor, the context "Import markdown" isn't quite intuitive. The label "Import markdown file" on the other side is more specific.

src/components/editor/document-bar/import/import-file.tsx
```diff
@@ -17,19 +17,26 @@ export const ImportFile: React.FC<ImportProps> = ({ noteContent, updateNoteConte
...
       const fileReader = new FileReader()
-      fileReader.readAsText(file)
       fileReader.addEventListener('load', () => {
         const newContent = fileReader.result as string
-        updateNoteContent(noteContent + '\n' + newContent)
+        if (noteContent.length === 0) {
+          updateNoteContent(newContent)
+        } else {
+          updateNoteContent(noteContent + '\n' + newContent)
+        }
```
DerMolly, line 26: should an upload really be appended? I thought it should replace the current content and create a new revision.

ErikMichelson: I would differentiate between the API call `PUT /note ` and the editor UI. Afaik the other imports don't replace the whole note content either.

mrdrogdrog (approved):
Maybe we should think about moving the editor content into the redux state..

ErikMichelson:
That could be an idea, but that's out of scope for this PR. Do you want to raise an issue for that?

mrdrogdrog:
I didn't want you to solve this in this PR. That's my I approved it :yum:

DerMolly (approved):

PR #652, opened 2020-10-07
Multi-Threaded Rendering

### LINKED ISSUE #410: Use web worker
Currently everything is calculated in the main thread. Even if the code is async, it's execution is done in ONE thread. This is slooow. Heavy caluclations (like the markdown to html rendering) are blocking the UI.

We could use [Web Worker](https://caniuse.com/#feat=webworkers). They can do this tasks in the background and take some load from the UI.

DerMolly:
### Component/Part
markdown renderer

### Description
This PR moves the cpu heavy rendering to a web worker

### Steps

- [x] added implementation
- [ ] added / updated tests
- [ ] extended changelog

### Related Issue(s)
fixes #410


cypress/integration/documentTitle.spec.ts
```diff
@@ -26,7 +26,7 @@ describe('Document Title', () => {
 
     it('yaml metadata title, opengraph title and first heading', () => {
       cy.get('.CodeMirror textarea')
-        .type(`---\ntitle: ${title}\nopengraph:\n  title: False title\n{backspace}{backspace}---\n# a first title`)
+        .type(`--\ntitle: ${title}\nopengraph:\n  title: False title\n{backspace}{backspace}---\n# a first title`)
```
ErikMichelson, line 29: Why did you remove a hyphen here?

DerMolly: that seems to be a bug. I'll fix it.

DerMolly:
The last commit tried to include the parser-debuger again. This yields the error `ReferenceError: $RefreshReg$ is not defined`. Removing the import and usage of `MarkdownItParserDebugger` in `BasicMarkdownItConfigurator` fixes it. @ErikMichelson and I tried to locate the problematic part in `MarkdownItParserDebugger` but weren't able to.

Currently if everything else runs, scrolling calls `usePostMetaDataOnChange`, which needs fixing.

mrdrogdrog:
@ErikMichelson  @DerMolly 
Can we close this PR?
The performance of the current renderer is good enough. Not sure if this is worth it. Multi-Threading could decrease the performance.

PR #695, opened 2020-10-18
Use fira code in editor

### LINKED ISSUE #452: Add possibility to change editor font (with ligatures)
Actually it's more about ligatures (i don't know if CodiMD supports them), but as I have `Fira Code` installed on my system, i really would like to use it in the editor to benefit from the ligatures provided by this font.

mrdrogdrog:
### Component/Part
Editor

### Description
This PR adds fira code as the editor font to support ligatures.

- [x] add font
- [x] Add editor preference for ligatures.

### Steps

- [x] added implementation
- [x] extended changelog

### Related Issue(s)
Close #452


CHANGELOG.md
```diff
@@ -70,6 +70,7 @@ SPDX-License-Identifier: CC-BY-SA-4.0
 - Use KaTeX instead of MathJax. ([Why?](https://community.codimd.org/t/frequently-asked-questions/190))
 - The dark-mode is also applied to the read-only-view and can be toggled from there.
 - Access tokens for the CLI and 3rd-party-clients can be managed in the user profile.
+- Change editor font to "Fira Code"
```
DerMolly, line 73: also adding the option to toggle ligatures

DerMolly (changes requested):

CHANGELOG.md
mrdrogdrog: done

src/components/editor/editor-pane/tool-bar/editor-preferences/editor-preference-ligatures-select.tsx
```diff
@@ -0,0 +1,26 @@
...
+  const saveLigatures = useCallback((event: ChangeEvent<HTMLSelectElement>) => {
+    const ligaturesActivated: boolean = event.target.value === 'true'
+    setEditorLigatures(ligaturesActivated)
+  }, [])
+
+  return (
+    <EditorPreferenceInput onChange={saveLigatures} value={ligaturesEnabled} property={"ligatures"}
+                           type={EditorPreferenceInputType.BOOLEAN}>
+      <option value='false'>Off</option>
+      <option value='true'>On</option>
```
DerMolly, line 23: Shouldn't this be translated

DerMolly (changes requested):

DerMolly (approved):

public/locales/en.json
```diff
@@ -389,11 +389,31 @@
...
+                },
                 "indentUnit": "Tab size (when using spaces)",
-                "spellcheck": "Spell checking"
+                "spellcheck": {
+                    "label": "Spell checking"
+                },
+                "ligatures": {
+                    "label": "Show ligatures",
+                    "on": "On",
+                    "off": "Off"
```
ErikMichelson, line 415: Aren't these common terms that should be placed under the `common.`-namespace?

mrdrogdrog: The words "on" and "off" yes. But in this case they are used in a special context. We could also write "Yoink" and "Yeet".

Also I don't want to add an exception to the code.

ErikMichelson: Yes, but we should minimize the required translation maintenance operations. When we have multiple on/off, yes/no, etc. translations, we could come to the point that these terms are not the same through the application. And if we want to manually maintain those cases, we'd have a lot of extra work.

Therefore I ask that you either add an exception to your code or rename these to "Ligatures enabled/disabled" or similar.

public/locales/en.json
```diff
@@ -389,11 +389,31 @@
...
+                    "sublime": "Sublime",
+                    "emacs": "Emacs",
+                    "vim": "Vim"
+                },
+                "indentWithTabs": {
+                    "label": "Tab character",
+                    "on": "Tabs",
+                    "off": "Spaces"
+                },
                 "indentUnit": "Tab size (when using spaces)",
```
ErikMichelson, line 408: Maybe the "when using spaces" could be removed as you wrapped the quite nice show-if around this field.

src/components/editor/editor-pane/tool-bar/editor-preferences/editor-preference-ligatures-select.tsx
```diff
@@ -0,0 +1,28 @@
...
+    setEditorLigatures(ligaturesActivated)
+  }, [])
+  const { t } = useTranslation()
+
+  return (
+    <EditorPreferenceInput onChange={saveLigatures} value={ligaturesEnabled} property={"ligatures"}
+                           type={EditorPreferenceInputType.BOOLEAN}>
+      <option value='false'>{t(`editor.modal.preferences.ligatures.off`)}</option>
+      <option value='true'>{t(`editor.modal.preferences.ligatures.on`)}</option>
+    </EditorPreferenceInput>
```
ErikMichelson, line 26: Why don't you use an `EditorPreferenceBooleanProperty` instead?

ErikMichelson (changes requested):

ErikMichelson:
In general: Why do you put ligatures as a separate option instead of moving that into the editor-config?
image

mrdrogdrog:
Because `preferences` only takes settings that are defined in the code mirror preference type

src/components/editor/editor-pane/tool-bar/editor-preferences/editor-preference-ligatures-select.tsx
mrdrogdrog: Because the property components don't offer an "onChange". They save the result into the redux state. I think this is okay until we add the spell check.

public/locales/en.json
mrdrogdrog: I changed the translation

ErikMichelson:
Okay, makes sense.

src/components/editor/editor-pane/tool-bar/editor-preferences/editor-preference-ligatures-select.tsx
ErikMichelson: Hm, maybe we should add an optional `onChange`-callback`?
But if you just want to do it like this, it's okay for me too.

mrdrogdrog: I don't think that it's worth here, because this is only a placeholder until the type checker is built.

src/redux/editor/reducers.ts
```diff
@@ -48,10 +50,18 @@ export const EditorConfigReducer: Reducer<EditorConfig, EditorConfigActions> = (
...
       saveToLocalStorage(newState)
       return newState
-    case EditorConfigActionType.SET_EDITOR_PREFERENCES:
+    case EditorConfigActionType.SET_LIGATURES:
       newState = {
         ...state,
-        preferences: (action as SetEditorPreferencesAction).preferences
+        ligatures: (action as SetEditorLigaturesAction).ligatures
+      }
+      return newState
```
ErikMichelson, line 58: You don't save this change to local-storage.

```suggestion
      }
      saveToLocalStorage(newState)
      return newState
```

ErikMichelson (changes requested):

public/locales/en.json
mrdrogdrog: I chose the exception. But not now. I already did it when I implemented the toggle :sweat_smile: . I forgot that the ligature toggle is a custom component and not a boolean property component.

src/redux/editor/reducers.ts
mrdrogdrog: done

src/components/editor/editor-pane/tool-bar/editor-preferences/editor-preference-ligatures-select.tsx
```diff
@@ -21,8 +21,8 @@ export const EditorPreferenceLigaturesSelect: React.FC = () => {
   return (
     <EditorPreferenceInput onChange={saveLigatures} value={ligaturesEnabled} property={"ligatures"}
                            type={EditorPreferenceInputType.BOOLEAN}>
-      <option value='false'>{t(`editor.modal.preferences.ligatures.off`)}</option>
-      <option value='true'>{t(`editor.modal.preferences.ligatures.on`)}</option>
+      <option value='false'>{t(`common.off`)}</option>
+      <option value='true'>{t(`common.on`)}</option>
```
ErikMichelson, line 25: Better but this time you're refering to `common.on` instead of `common.yes`. For no/off similar.

mrdrogdrog: changed

src/components/editor/editor-pane/tool-bar/editor-preferences/editor-preference-ligatures-select.tsx
```diff
@@ -21,8 +21,8 @@ export const EditorPreferenceLigaturesSelect: React.FC = () => {
   return (
     <EditorPreferenceInput onChange={saveLigatures} value={ligaturesEnabled} property={"ligatures"}
                            type={EditorPreferenceInputType.BOOLEAN}>
-      <option value='false'>{t(`common.off`)}</option>
-      <option value='true'>{t(`common.on`)}</option>
+      <option value='false'>{t(`common.yes`)}</option>
+      <option value='true'>{t(`common.no`)}</option>
```
ErikMichelson, line 25: Yes and no are swapped.

ErikMichelson (approved):

mrdrogdrog:
FINALLY

Issue #718, opened 2019-11-30
Number of character / number of words


parisni:
hello

intersting information are:
- number of charcacter in the resulting format (html, odm, pdf) 
- number of words in the resulting format

right now the information of number of charracter is about every single markdown markup

ErikMichelson:
For the word-count this library looks quite complete in regard to UTF-8 character word-recognition: https://www.npmjs.com/package/words-count

DerMolly:
The request is concerned with the word and character count in the result. So currently only HTML, because there is no PDF export at this moment. This could be done with the html output, but we would need something that extracts the correct word & character count out of an HTML document.

In the mean time we could add word count via the lib @ErikMichelson suggested to the infos in the status bar below the editor. Between lines and length (aka characters would be a neat place)
![grafik](https://user-images.githubusercontent.com/3978662/97745856-c9c89500-1ae9-11eb-9fa0-7922a43acc80.png)

ErikMichelson:
This has been implemented

PR #738, opened 2020-11-02
Add word count in document info modal

ErikMichelson:
### Component/Part
Editor -> Status bar

### Description
This PR adds the count of words in the current document to the document info dialog.
The method counts the words in the rendered version of the document excluding image captions and codeblocks (only full blocks are ignored, inline is counted). As counting algorithm itself the library [word-count](https://www.npmjs.com/package/word-count) is taken.

### Screenshots
image

### Steps

- [x] added implementation
- [x] added / updated tests
- [x] extended changelog

### Related Issue(s)
#718

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -73,6 +73,7 @@ SPDX-License-Identifier: CC-BY-SA-4.0
 - When pasting tables (e.g. from LibreOffice Calc or MS Excel) they get reformatted to markdown tables.
 - The history page supports URL parameters that allow bookmarking of a specific search of tags filter.
 - Users can change the pinning state of a note directly from the editor.
+- Count of words in final document is viewable via new document info dialog.
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -73,7 +73,7 @@ SPDX-License-Identifier: CC-BY-SA-4.0
 - When pasting tables (e.g. from LibreOffice Calc or MS Excel) they get reformatted to markdown tables.
 - The history page supports URL parameters that allow bookmarking of a specific search of tags filter.
 - Users can change the pinning state of a note directly from the editor.
-- Count of words in final document is viewable via new document info dialog.
+- Note information dialog containing word count, revision count, last editor and creation time.
 
 ### Changed
 
--- a/cypress/integration/word-count.spec.ts
+++ b/cypress/integration/word-count.spec.ts
@@ -12,32 +12,32 @@ describe('Test word count with', () => {
   it('empty note', () => {
     cy.codemirrorFill('')
     cy.wait(500)
-    cy.get('button[data-cy="sidebar-btn-document-info"]').click()
+    cy.get('[data-cy="sidebar-btn-document-info"]').click()
     cy.get('[data-cy="document-info-modal"]').should('be.visible')
-    cy.get('b[data-cy="document-info-word-count"]').should('have.text', '0')
+    cy.get('[data-cy="document-info-word-count"]').should('have.text', '0')
   })
 
   it('simple words', () => {
     cy.codemirrorFill('five words should be enough')
     cy.wait(500)
-    cy.get('button[data-cy="sidebar-btn-document-info"]').click()
+    cy.get('[data-cy="sidebar-btn-document-info"]').click()
     cy.get('[data-cy="document-info-modal"]').should('be.visible')
-    cy.get('b[data-cy="document-info-word-count"]').should('have.text', '5')
+    cy.get('[data-cy="document-info-word-count"]').should('have.text', '5')
   })
 
   it('excluded codeblocks', () => {
     cy.codemirrorFill('```\nthis is should be ignored\n```\n\ntwo `words`')
     cy.wait(500)
-    cy.get('button[data-cy="sidebar-btn-document-info"]').click()
+    cy.get('[data-cy="sidebar-btn-document-info"]').click()
     cy.get('[data-cy="document-info-modal"]').should('be.visible')
-    cy.get('b[data-cy="document-info-word-count"]').should('have.text', '2')
+    cy.get('[data-cy="document-info-word-count"]').should('have.text', '2')
   })
 
   it('excluded images', () => {
     cy.codemirrorFill('![ignored alt text](https://dummyimage.com/48) not ignored text')
     cy.wait(500)
-    cy.get('button[data-cy="sidebar-btn-document-info"]').click()
+    cy.get('[data-cy="sidebar-btn-document-info"]').click()
     cy.get('[data-cy="document-info-modal"]').should('be.visible')
-    cy.get('b[data-cy="document-info-word-count"]').should('have.text', '3')
+    cy.get('[data-cy="document-info-word-count"]').should('have.text', '3')
   })
 })
--- a/cypress/integration/word-count.spec.ts
+++ b/cypress/integration/word-count.spec.ts
@@ -1,0 +1,43 @@
+/*
+ * SPDX-FileCopyrightText: 2021 The HedgeDoc developers (see AUTHORS file)
+ *
+ * SPDX-License-Identifier: AGPL-3.0-only
+ */
+
+describe('Test word count with', () => {
+  beforeEach(() => {
+    cy.visitTestEditor()
+  })
+
+  it('empty note', () => {
+    cy.codemirrorFill('')
+    cy.wait(500)
+    cy.get('button[data-cy="sidebar-btn-document-info"]').click()
+    cy.get('[data-cy="document-info-modal"]').should('be.visible')
+    cy.get('b[data-cy="document-info-word-count"]').should('have.text', '0')
+  })
+
+  it('simple words', () => {
+    cy.codemirrorFill('five words should be enough')
+    cy.wait(500)
+    cy.get('button[data-cy="sidebar-btn-document-info"]').click()
+    cy.get('[data-cy="document-info-modal"]').should('be.visible')
+    cy.get('b[data-cy="document-info-word-count"]').should('have.text', '5')
+  })
+
+  it('excluded codeblocks', () => {
+    cy.codemirrorFill('```\nthis is should be ignored\n```\n\ntwo `words`')
+    cy.wait(500)
+    cy.get('button[data-cy="sidebar-btn-document-info"]').click()
+    cy.get('[data-cy="document-info-modal"]').should('be.visible')
+    cy.get('b[data-cy="document-info-word-count"]').should('have.text', '2')
+  })
+
+  it('excluded images', () => {
+    cy.codemirrorFill('![ignored alt text](https://dummyimage.com/48) not ignored text')
+    cy.wait(500)
+    cy.get('button[data-cy="sidebar-btn-document-info"]').click()
+    cy.get('[data-cy="document-info-modal"]').should('be.visible')
+    cy.get('b[data-cy="document-info-word-count"]').should('have.text', '3')
+  })
+})
--- a/public/locales/en.json
+++ b/public/locales/en.json
@@ -347,7 +347,8 @@
                 "created": "<0></0> created this note <1></1>",
                 "edited": "<0></0> was the last editor <1></1>",
                 "usersContributed": "<0></0> users contributed to this document",
-                "revisions": "<0></0> revisions are saved"
+                "revisions": "<0></0> revisions are saved",
+                "words": "<0></0> words in document"
             },
             "gistImport": {
                 "title": "Import from Gist",
@@ -458,6 +459,7 @@
         "and": "and",
         "avatarOf": "avatar of '{{name}}'",
         "why": "Why?",
+        "loading": "Loading ...",
         "successfullyCopied": "Copied!",
         "copyError": "Error while copying!",
         "errorOccurred": "An error occurred",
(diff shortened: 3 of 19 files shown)


public/locales/en.json
```diff
@@ -306,6 +306,7 @@
                 "line": "Selected {{count}} lines"
             },
             "lines": "{{lines}} Lines",
+            "words": "{{words}} Words",
```
SISheogorath, line 309: Do we handle plurals here?

src/components/editor/editor-pane/status-bar/status-bar.tsx
```diff
@@ -27,11 +30,20 @@ export const createStatusInfo = (editor: Editor, maxDocumentLength: number): Sta
   charactersInDocument: editor.getValue().length,
   remainingCharacters: maxDocumentLength - editor.getValue().length,
   linesInDocument: editor.lineCount(),
+  wordsInDocument: wordsCount(editor.getValue()),
```
SISheogorath, line 33: So how accurate is this? How does it handle code blocks? How does it handle links? How does it handle raw HTML?

SISheogorath (commented):
Mainly some questions about the feature

src/components/editor/editor-pane/status-bar/status-bar.tsx
mrdrogdrog: The lib is not code specific. So it would count ANYTHING, that is separated by a space, which is useless imho.

The only usage that could be useful is, if we count the text-only content from the render output.

ErikMichelson: Well actually it does not split only at spaces but recognizes CJK words etc. But it's right, that approach is counting every word in the markdown "code".

public/locales/en.json
ErikMichelson: I didn't do that here as we didn't do that at many other places as well. Maybe that's a thing for a separate issue?

src/components/editor/editor-pane/status-bar/status-bar.tsx
ErikMichelson: Counting words in the HTML render output is a thing that we definitely should do in the long term, but I would rather locate such information into the "document statistics"-modal than in the editor-statusbar.

SISheogorath: That's basically the problem I mentioned in the related issue. There are problems ranging from people commenting large parts of the text and the counter would actually increase not decrease, and even when we used the rendered output it's quite likely that it would still count code blocks with it.

It's a wrong counter and I'm not sure if that is really helpful. If we really want to integrate it, we should make it clear, that this is a rough estimation (e.g. by adding a `~` in front of the number) as I can already see lots of bug reports flowing in, telling us that our counter is wrong (e.g. "Simple sentence with smiley at the end is counted wrong. There are 5 not 6 words. Example: `This is a simple sentence. :smile:`").

ErikMichelson:
To touch this issue again:
We should probably drop the word count based on the editor content (as done here), reuse some code and instead count words in the HTML output that aren't inside multi-line code blocks.
The information probably fits better inside the document statistics modal.

For now this is therefore waiting for the rendering rewrite (iframe based) to get the words out of that.

DerMolly (approved):

CHANGELOG.md
```diff
@@ -73,6 +73,7 @@ SPDX-License-Identifier: CC-BY-SA-4.0
 - When pasting tables (e.g. from LibreOffice Calc or MS Excel) they get reformatted to markdown tables.
 - The history page supports URL parameters that allow bookmarking of a specific search of tags filter.
 - Users can change the pinning state of a note directly from the editor.
+- Count of words in final document is viewable via new document info dialog.
```
mrdrogdrog, line 76: This sounds like "the info dialog for new documents". How about "Add word count to document information dialog," ?

cypress/integration/word-count.spec.ts
```diff
@@ -0,0 +1,43 @@
...
+  beforeEach(() => {
+    cy.visitTestEditor()
+  })
+
+  it('empty note', () => {
+    cy.codemirrorFill('')
+    cy.wait(500)
+    cy.get('button[data-cy="sidebar-btn-document-info"]').click()
+    cy.get('[data-cy="document-info-modal"]').should('be.visible')
+    cy.get('b[data-cy="document-info-word-count"]').should('have.text', '0')
```
mrdrogdrog, line 17: you don't need that b. The tag selector should be enough.

src/components/editor-page/document-bar/document-info/document-info-line-word-count.tsx
```diff
@@ -0,0 +1,40 @@
...
+import { DocumentInfoLine } from './document-info-line'
+import { UnitalicBoldText } from './unitalic-bold-text'
+import { useIFrameEditorToRendererCommunicator } from '../../render-context/iframe-editor-to-renderer-communicator-context-provider'
+
+export const DocumentInfoLineWordCount: React.FC = () => {
+  useTranslation()
+  const iframeEditorToRendererCommunicator = useIFrameEditorToRendererCommunicator()
+
+  // -1 means the word count isn't determined yet, any non-negative number is the actual counting result
+  const [wordCount, setWordCount] = useState(-1)
```
mrdrogdrog, line 19: 1. Magic number. Use a const for that.
2. Why don't you use undefined or null for that?

src/components/editor-page/document-bar/document-info/document-info-line-word-count.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+  useEffect(() => {
+    iframeEditorToRendererCommunicator?.onWordCountCalculated((words) => {
+      setWordCount(words)
+    })
```
mrdrogdrog, line 24: you don't unbind the handler when the effect ends. Just set it to undefined.

src/components/editor-page/document-bar/document-info/document-info-line-word-count.tsx
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+import React, { useEffect, useState } from 'react'
+import { Trans, useTranslation } from 'react-i18next'
+import { ShowIf } from '../../../common/show-if/show-if'
+import { DocumentInfoLine } from './document-info-line'
+import { UnitalicBoldText } from './unitalic-bold-text'
+import { useIFrameEditorToRendererCommunicator } from '../../render-context/iframe-editor-to-renderer-communicator-context-provider'
+
+export const DocumentInfoLineWordCount: React.FC = () => {
```
mrdrogdrog, line 14: I started to add js docs to my components. Maybe you want to jump on that train, too? :yum:

src/components/render-page/iframe-markdown-renderer.tsx
```diff
@@ -22,10 +23,24 @@ export const IframeMarkdownRenderer: React.FC = () => {
 
   const iframeCommunicator = useIFrameRendererToEditorCommunicator()
 
+  const countWordsInRenderedDocument = useCallback(() => {
+    const documentContainer = document.querySelector('.markdown-body')
+    if (!documentContainer) {
+      iframeCommunicator?.sendWordCountCalculated(0)
+      return
+    }
+    const words = countWordsRecursive(documentContainer)
```
mrdrogdrog, line 32:
```suggestion
    const wordCount = countWordsRecursive(documentContainer)
```

src/components/render-page/word-counter.ts
```diff
@@ -0,0 +1,28 @@
...
+ * SPDX-FileCopyrightText: 2021 The HedgeDoc developers (see AUTHORS file)
+ *
+ * SPDX-License-Identifier: AGPL-3.0-only
+ */
+
+import wordsCount from 'words-count'
+
+const EXCLUDED_TAGS = ['img', 'pre', 'nav']
+
+export const countWordsRecursive = (node: Element | ChildNode): number => {
```
mrdrogdrog, line 11: Is it important for a caller to know that this method works recursivly? I don't see whats wrong with just "countWords".

src/components/render-page/word-counter.ts
```diff
@@ -0,0 +1,28 @@
...
+  let words = 0
```
mrdrogdrog, line 12: Why this var? You use it only in one case. And this case could be replaced with a `.reduce`

src/components/render-page/word-counter.ts
```diff
@@ -0,0 +1,28 @@
...
+  if (
+    !node ||
```
mrdrogdrog, line 14: how can the node be false?

src/components/render-page/word-counter.ts
```diff
@@ -0,0 +1,28 @@
...
+    (node as HTMLElement).classList?.contains('katex-mathml')
+  ) {
+    return 0
+  }
+  if (node.hasChildNodes()) {
+    node.childNodes.forEach((childNode) => {
+      words += countWordsRecursive(childNode)
+    })
+  } else {
+    return wordsCount(node.textContent ?? '')
```
mrdrogdrog, line 25: No text content doesn't contain any words, right? I would put the case "no text content" into a seperate if branch :thinking:

src/components/render-page/word-counter.ts
```diff
@@ -0,0 +1,28 @@
...
+import wordsCount from 'words-count'
+
+const EXCLUDED_TAGS = ['img', 'pre', 'nav']
+
+export const countWordsRecursive = (node: Element | ChildNode): number => {
+  let words = 0
+  if (
+    !node ||
+    EXCLUDED_TAGS.includes(node.nodeName.toLowerCase()) ||
+    (node as HTMLElement).classList?.contains('katex-mathml')
```
mrdrogdrog, line 16: Can you extract both of these conditions into methods that say what you want to achieve with this conditons? Like `is ignored html tag`

src/components/render-page/word-counter.ts
mrdrogdrog, line 11: docs would be nice

mrdrogdrog (changes requested):

src/components/render-page/word-counter.ts
ErikMichelson: I'd say it's better for code understanding if someone can quickly determine that this method is called recursively.

mrdrogdrog: The name is more important for the person that users the function and not the person that is reviewing it. The name should describe the purpose of the function and not how it works.

CHANGELOG.md
ErikMichelson: changed

cypress/integration/word-count.spec.ts
ErikMichelson: fixed

src/components/editor-page/document-bar/document-info/document-info-line-word-count.tsx
ErikMichelson: changed

src/components/editor-page/document-bar/document-info/document-info-line-word-count.tsx
ErikMichelson: done

ErikMichelson: However, we should create a PR adding specific ESDoc conventions for our project (for example in the contribution guide).

src/components/render-page/iframe-markdown-renderer.tsx
ErikMichelson: changed

src/components/render-page/word-counter.ts
ErikMichelson: changed if you insist on it

src/components/render-page/word-counter.ts
ErikMichelson: changed

src/components/render-page/word-counter.ts
ErikMichelson: Maybe a little `false` goes astray and is caught by this if?
Joke aside, changed it.

src/components/render-page/word-counter.ts
ErikMichelson: changed

src/components/render-page/word-counter.ts
ErikMichelson: changed

src/components/render-page/word-counter.ts
ErikMichelson: done

src/components/render-page/word-counter.ts
```diff
@@ -0,0 +1,50 @@
...
+ * @return The number of words counted in this node and its children.
+ */
+export const countWords = (node: Element | ChildNode): number => {
+  if (!node.textContent || isExcludedTag(node) || isExcludedClass(node)) {
+    return 0
+  }
+  if (!node.hasChildNodes()) {
+    return wordsCount(node.textContent)
+  }
+  return [...node.childNodes].reduce((words, childNode) => {
```
mrdrogdrog, line 47: Why are you using the spread syntax here? :thinking: Is there a specific reason?

mrdrogdrog (approved):

src/components/render-page/word-counter.ts
ErikMichelson: `node.childNodes` is a DOMNodeList and not an array.

DerMolly (approved):

PR #793, opened 2020-11-26
Replace source * pro files with npm package

mrdrogdrog:
### Component/Part
Repository structure

### Description
This PR deletes the font files of "Source Sans Pro" (and others) from the repository and adds a npm package, that provides these fonts.

### Steps

- [x] Added implementation


DerMolly (approved):

DerMolly:
I though that we only needed the `OFL-1.1-RFN` License for these files and wanted to suggest to remove the License then, too, but it seems the license is used else where. At least reuse lint does not mention any unused licenses. But I can't find any files with this license… Maybe we found a bug in the reuse tool?

mrdrogdrog:
Seems so. Let's see what reuse says when this is merged.

ErikMichelson (approved):

DerMolly (approved):

PR #860, opened 2020-12-17
Enhance share dialog

ErikMichelson:
### Component/Part
Editor → document bar → share button

### Description
This PR improves the dialog that opens after clicking the share link button. It now includes two share-able URLs: One for the editor in the current view-mode and one for the read-only-view or presentation-view depending on the note type.

### Steps

- [x] Added implementation
- [x] Extended changelog
- [x] I read the [contribution documentation](https://github.com/hedgedoc/react-client/blob/main/CONTRIBUTING.md) and signed-off my commits to accept the DCO.

### Related Issue(s)
Fixes #790 
Fixes #791

PR #873, opened 2020-12-22
Add support for editor-mode URL parameters

ErikMichelson:
### Component/Part
Editor

### Description
This PR adds support for respecting the editor-mode URL parameters (?both, ?edit, ?view).

When testing this branch, please note that the editor-mode enum has changed and therefore the localStorage-stored-settings won't work anymore unless you once click on one of the edit-view-both buttons once.

### Steps

- [x] Added implementation
- [x] Added / updated tests

### Related Issue(s)
#860

### DIFF
--- a/src/components/editor/editor.tsx
+++ b/src/components/editor/editor.tsx
@@ -7,7 +7,7 @@ SPDX-License-Identifier: AGPL-3.0-only
 import React, { Fragment, useCallback, useEffect, useRef, useState } from 'react'
 import { useTranslation } from 'react-i18next'
 import { useSelector } from 'react-redux'
-import { useParams } from 'react-router'
+import { useLocation, useParams } from 'react-router'
 import useMedia from 'use-media'
 import { useApplyDarkMode } from '../../hooks/common/use-apply-dark-mode'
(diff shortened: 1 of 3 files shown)


src/components/editor/editor.tsx
```diff
@@ -60,7 +61,12 @@ export const Editor: React.FC = () => {
 
   useEffect(() => {
     setDocumentContent(editorTestContent)
-  }, [])
+    const requestedMode = search.substr(1)
+    const mode = Object.values(EditorMode).filter(mode => mode === requestedMode)
```
mrdrogdrog, line 65: Use find instead of filter

mrdrogdrog (changes requested):

src/components/editor/editor.tsx
ErikMichelson: Good point, I just forgot about that method. Will change that soon.

ErikMichelson: Done

mrdrogdrog (changes requested):
Can you try to get rid of this "isFirstDraw" hack?

mrdrogdrog (approved):

DerMolly (approved):

Issue #103, opened 2020-05-31
Offline support


ErikMichelson:
**Is your feature request related to a problem? Please describe.**
No

**Describe the solution you'd like**
While being in offline mode, you still may create notes which will be stored by the service worker on your local device and that will be uploaded to the backend server the next time you have an internet connection.  
Optionally: Pinned notes will be downloaded and cached regularly by the service worker so that you can read (not edit!) them when the device is offline.

**Describe alternatives you've considered**
The alternative would be to have absolutely no offline support like the current version.

PR #896, opened 2020-12-27
Register service worker in production

mrdrogdrog:
### Component/Part
Service worker

### Description
This PR activates the service worker for production builds.

Please review https://github.com/hedgedoc/react-client/pull/898 first!

### Steps

- [x] Added implementation
- [x] Extended changelog

### Related Issue(s)
#103

### DIFF
--- a/public/locales/en.json
+++ b/public/locales/en.json
@@ -455,6 +455,15 @@
         "copyError": "Error while copying!",
         "errorOccurred": "An error occurred"
     },
+    "global": {
+        "modal": {
+            "reloadApp": {
+                "title": "Update available",
+                "content": "An update for hedgedoc is available. Please reload the page.",
+                "reloadButton": "Reload app"
+            }
+        }
+    },
     "login": {
         "chooseMethod": "Choose method",
         "signInVia": "Sign in via {{service}}",
--- a/public/locales/en.json
+++ b/public/locales/en.json
@@ -459,7 +459,7 @@
         "modal": {
             "reloadApp": {
                 "title": "Update available",
-                "content": "An update for hedgedoc is available. Please reload the page.",
+                "content": "An update for hedgedoc is available. Please reload every hedgedoc tab.",
                 "reloadButton": "Reload app"
             }
         }
--- a/public/locales/en.json
+++ b/public/locales/en.json
@@ -459,7 +459,7 @@
         "modal": {
             "reloadApp": {
                 "title": "Update available",
-                "content": "An update for hedgedoc is available. Please reload every hedgedoc tab.",
+                "content": "An update for HedgeDoc is available. Please reload every HedgeDoc tab.",
                 "reloadButton": "Reload app"
             }
         }
--- a/src/service-worker-registration.ts
+++ b/src/service-worker-registration.ts
@@ -32,37 +32,36 @@ type Config = {
 };
 
 export function register (config?: Config): void {
-  if (process.env.NODE_ENV === 'production' && 'serviceWorker' in navigator) {
-    // The URL constructor is available in all browsers that support SW.
-    const publicUrl = new URL(process.env.PUBLIC_URL, window.location.href)
-    if (publicUrl.origin !== window.location.origin) {
-      // Our service worker won't work if PUBLIC_URL is on a different origin
-      // from what our page is served on. This might happen if a CDN is used to
-      // serve assets; see https://github.com/facebook/create-react-app/issues/2374
-      return
-    }
-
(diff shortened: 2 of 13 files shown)


DerMolly:
How does this get triggered? Currently it won't, or am I missing something?

mrdrogdrog:
What do you mean?

DerMolly (approved):

mrdrogdrog:
Okay now you can review this.

DerMolly (approved):

public/locales/en.json
```diff
@@ -455,6 +455,15 @@
         "copyError": "Error while copying!",
         "errorOccurred": "An error occurred"
     },
+    "global": {
+        "modal": {
+            "reloadApp": {
+                "title": "Update available",
+                "content": "An update for hedgedoc is available. Please reload every hedgedoc tab.",
```
ErikMichelson, line 462:
```suggestion
                "content": "An update for HedgeDoc is available. Please reload every HedgeDoc tab.",
```

src/service-worker-registration.ts
```diff
@@ -115,36 +114,33 @@ function checkValidServiceWorker (swUrl: string, config?: Config) {
...
-      // Service worker found. Proceed as normal.
+        // Service worker found. Proceed as normal.
         registerValidSW(swUrl, config)
       }
     })
-    .catch(() => {
-      console.log('No internet connection found. App is running in offline mode.')
-    })
+  .catch(() => {
+    console.log('No internet connection found. App is running in offline mode.')
```
ErikMichelson, line 132: Do we want to store a field `offline` in the redux (for example in the new notifications-object)? This way we could later show a "You're in offline-mode. Live-changes won't be applied."-message.

src/service-worker-registration.ts
```diff
@@ -77,22 +76,22 @@ function registerValidSW (swUrl: string, config?: Config) {
...
-            // but the previous service worker will still serve the older
-            // content until all client tabs are closed.
+              // At this point, the updated precached content has been fetched,
+              // but the previous service worker will still serve the older
+              // content until all client tabs are closed.
               console.log(
                 'New content is available and will be used when all ' +
-              'tabs for this page are closed. See https://cra.link/PWA.'
+                'tabs for this page are closed. See https://cra.link/PWA.'
               )
```
ErikMichelson, line 85: Customize or remove this message?
(For me it's not about the message but about the cra.link which could instead point to our FAQ)

src/service-worker-registration.ts
```diff
@@ -32,37 +32,36 @@ type Config = {
...
-      }
-    })
-  }
+      // Add some additional logging to localhost, pointing developers to the
+      // service worker/PWA documentation.
+      navigator.serviceWorker.ready.then(() => {
+        console.log(
+          'This web app is being served cache-first by a service ' +
+          'worker. To learn more, visit https://cra.link/PWA'
+        )
```
ErikMichelson, line 58: Same as with the other comment on cra.link I've already written.

public/locales/en.json
```diff
@@ -455,6 +455,15 @@
...
+                "reloadButton": "Reload app"
```
ErikMichelson, line 463: This in total is quite similar to another message we already have: `editor.error.newVersion`. To reduce the load on the translators, maybe we could use that instead or merge them?

ErikMichelson (changes requested):
Although the service-worker itself works, the offline-support it should bring doesn't as the ApplicationInitializer fails to fetch the content from online-server. We should therefore probably cache the api/config as well.

src/service-worker-registration.ts
mrdrogdrog: I removed this message, because we only need the callback.

public/locales/en.json
mrdrogdrog: I would remove it becaus its no error and we have a new modal for that.

public/locales/en.json
mrdrogdrog: done

public/locales/en.json
ErikMichelson: Should we move the already translated strings to the new message if applicable?

mrdrogdrog: It's another translation. And tbh I don't like the old translation.

mrdrogdrog:
Maybe we should delete this branch and rewrite the service worker as soon as we know what we want to cache offline. Like API stuff

ErikMichelson:
But why should we delete it then? Seems like we could extend it later.

PR #897, opened 2020-12-27
Rename npm task

mrdrogdrog:
### Component/Part
NPM task "start:dev"

### Description
IMHO it's very confusing, that there is a task "start" and a task "start:dev", which is for usage with a backend. Therefore I propose to rename the backend task

### Steps

- [x] Added implementation

### DIFF
--- a/package.json
+++ b/package.json
@@ -108,7 +108,7 @@
     },
     "scripts": {
         "start": "PORT=3001 craco start",
-        "start:for-backend": "REACT_APP_BACKEND=http://localhost:3000 yarn start",
+        "start:for-real-backend": "REACT_APP_BACKEND=http://localhost:3000 yarn start",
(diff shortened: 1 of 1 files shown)


ErikMichelson:
Indeed the current situation is bad and I'm not opposed to your suggestion.
Just to make the task names even more clear, what about the following?
 - `start` -> `start:with-mocking`
 - `start:dev` -> `start:with-backend`

mrdrogdrog:
We can do that, but I won't delete the `start` task, because it is a convetional default task. Maybe we should just alias it?

mrdrogdrog:
Also: `start:with-mock` doesn't enable or disable anything. It doesn't change anything except the port. Therefore `start:for-backend` is an alternative use case. I don't see a problem with using just `start` if we don't change anything.

DerMolly (approved):

ErikMichelson (approved):

PR #915, opened 2021-01-06
Fix yaml metadata and document title not resetting on deletion

ErikMichelson:
### Component/Part
redux states and title extraction logic

### Description
This PR fixes the fact that yaml-metadata as well as the first heading for the title are cached for too long even if the document has no content anymore.

### Steps

- [x] Added implementation

### DIFF
--- a/src/components/markdown-renderer/basic-markdown-renderer.tsx
+++ b/src/components/markdown-renderer/basic-markdown-renderer.tsx
@@ -23,6 +23,7 @@ export interface BasicMarkdownRendererProps {
   markdownIt: MarkdownIt,
   documentReference?: RefObject<HTMLDivElement>
   onBeforeRendering?: () => void
+  onPostRendering?: () => void
 }
 
 export const BasicMarkdownRenderer: React.FC<BasicMarkdownRendererProps & AdditionalMarkdownRendererProps> = ({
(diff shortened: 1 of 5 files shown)


mrdrogdrog (changes requested):
There is still a bug that causes the "yaml data invalid" message to stay if I remove a document with an invalid frontmatter

ErikMichelson:
We update the error boolean in the callback of the markdown-it-frontmatter plugin. It seems that this callback is never fired when there's no frontmatter thus the boolean doesn't get updated. I'm currently looking into how to resolve this.

mrdrogdrog:
maybe we should merge https://github.com/hedgedoc/react-client/pull/916 before this pr?

ErikMichelson:
As the markdown-it-frontmatter plugins runs on every content change of the whole markdown-content, we could disable this boolean on every content change via a useEffect prior to the plugin running. Hm, sounds a little hacky but maybe worth trying?

ErikMichelson:
Yes

mrdrogdrog:
Set yamlerror 

> As the markdown-it-frontmatter plugins runs on every content change of the whole markdown-content, we could disable this boolean on every content change via a useEffect prior to the plugin running. Hm, sounds a little hacky but maybe worth trying?

Well. How about this: When rendering a document we assume that there are no errors. If there are errors (the callback will get fired) we set a boolean (a ref or var) to true. After the rendering we check if the value is different from the current state and change it if necessary.

mrdrogdrog:
wow. deja vu. Don't we do that already somewhere? o.ö

ErikMichelson:
Isn't that the same thing I meant but just a bit more complex?

mrdrogdrog:
Don't do unnecessary state changes. Every state change triggers a rerender!

ErikMichelson:
You're right with that, I forgot that we need to use a state there.

mrdrogdrog (approved):

DerMolly (approved):

PR #918, opened 2021-01-06
Feature/placeholder image

### LINKED ISSUE #406: Feature request: Image placeholders
When typing an image element without any source URL, it would be helpful if a placeholder element could be displayed which when clicked allows the user to select an image for upload.

Example syntax: `![Placeholder text]`

DerMolly:
### Component/Part
markdown render

### Description
This PR adds placeholder images.

This still needs some work to make the image uploading work. That can be done after #837 is merged.

### Steps

- [ ] Added implementation

### Related Issue(s)
fixes #406


mrdrogdrog:
IMHO we should talk again about the placeholder syntax. Because "Link is invalid, so it must be a placeholder" is a bit unintuitive.

DerMolly:
It's just `https://`. It was an easy solution and as @ErikMichelson mentioned `https://` is what the buttons and autocompletions insert.

mrdrogdrog:
well. shouldn't we then edit the autocompletion instead? :D

src/components/markdown-renderer/replace-components/image/image-frame.tsx
```diff
@@ -23,10 +24,17 @@ export const ImageFrame: React.FC<React.ImgHTMLAttributes<HTMLImageElement>> = (
       .catch(err => console.error(err))
   }, [imageProxyEnabled, src])
 
+  if (src === 'https://') {
+    return (
+      <PlaceholderImageFrame alt={alt} title={title} width={props.width} height={props.height} />
+    )
+  }
```
mrdrogdrog, line 31: I'm not sure if I'm happy with this block. The purpose of the ImageFrame-component is to show a real image. Maybe the image replacer would be a better place?

DerMolly:
Not sure if I like this. Because the auto-completion and button insert `https://` as a simple start for users to insert an image. Also it's the behavior users are used to from 1.x releases.

mrdrogdrog:
This wouldn't be our first breaking change.

mrdrogdrog:
Well. How about a second autocompletion? Image with URL stub and Image with placeholder?

ErikMichelson:
I'm against a second autocompletion, that would just rise more confusion under the users. What's the problem about the `https://`? It's never a valid image link and allows users to quickly see that they also could insert a correct URL instead.

mrdrogdrog:
Because there is no intuitive connection between the text "https://" and "i'm showing a placeholder for that image". We're just using this text because our autocompletion had this stub before.

ErikMichelson:
I thought the intention is to show the placeholder when waiting for content. And this is just the string that should be there when waiting for content.

Look at it from a user's viewpoint: When they see an unfinished link there, they directly know where to paste the correct image URL even if they haven't used markdown before. A string "placeholder" is meaningless.

mrdrogdrog:
IMHO We would get the same effect if we use "<LINK GOES HERE>" (or similar)

DerMolly:
> Because there is no intuitive connection between the text "https://" and "i'm showing a placeholder for that image". We're just using this text because our autocompletion had this stub before.

I beg to differ. I think it's rather intuitive that there is custom magic word you need to know, but just a simple unfinished image link to trigger this.

But it seems we can't decide among us. Maybe we should ask the others devs? Or ask in the forum?

mrdrogdrog:
TL;DR

We're not sure which syntax is better:

`alttext`
with the literal "https://" string as link.

or

`alttext`
with the literal "<placeholder>" string as link.

(title and size are part of the default markdown image syntax and optional)

mrdrogdrog:
The first one would make it clear where to put the link. The second one would make it more obvious, that you're showing a placeholder.

InnayTool:
I prefer the first solution in combination with an extra image for broken links, with the same behaviour.

DerMolly:
We had a look at the handling of failed images and @ErikMichelson found this stackoverflow post https://stackoverflow.com/a/9891041.
We're thinking of handling the error by including the placeholder image frame with an additional optional prop warning (or something along those lines) to indicate that an image failed to load that was supposed to be there and still offer the option to upload an image.
The difference is important as an placeholder is by definition the absence of the correct image, but an missing image is an error and should be clearly communicated.

Normally this is would be the browsers domain, but not all the browser (firefox e.g.)  handle this in a great way.

mrdrogdrog:
Sry but I'm still not sure if this is a good idea :/.
I'm mean: okay, we can show a placeholder if an image can't be loaded.. but i don't think we should use the "this image is undefined, upload one here" placeholder

PR #935, opened 2021-01-09
Add branch info to browserstack runs

ErikMichelson:
### Component/Part
GitHub actions workflows -> E2E browserstack

### Description
This PR adds branch information to the browserstack runs to easier identify them in the browserstack dashboard.

As @mrdrogdrog noted, the BrowserStack setup-env GitHub action has a field for build-name and project-name, but several commits (273772d, 928d312, 2f287b9, 4d53819, 216a90d, 7432e16, 09a39d8, 9ac28c4) showed they aren't working as expected.

Regarding the git commit sha extraction:
GitHub runs the tests in PRs against a simulated merge-commit with the base-branch. The commit SHA that is outputted by either git itself or by `${{github.sha}}` therefore is useless. Luckily GitHub provides the SHA of the commit that triggered a PR-run in a separate variable `${{github.event.pull_request.head.sha}}`. This variable isn't set on pushes to main, so I included a little if-else to use either the frist or the second variable.

### Steps

- [x] Added implementation

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -18,6 +18,14 @@ jobs:
       - name: Check out repo
         uses: actions/checkout@v2
 
+      - name: Get branch and commit
+        id: gitinfo
+        run: |
+          if [ -n "${{ github.event.pull_request.head.sha }}" ]; \
+          then echo "::set-output name=shapr::$(echo ${{ github.event.pull_request.head.sha }} | cut -c -8) - #${{ github.event.pull_request.number }}"; \
+          else echo "::set-output name=shapr::$(echo ${{ github.sha }} | cut -c -8)"; fi
+          echo "::set-output name=branch::${GITHUB_HEAD_REF:-main}"
+
       - name: Get yarn cache directory path
         id: yarn-cache-dir-path
         run: echo "::set-output name=dir::$(yarn cache dir)"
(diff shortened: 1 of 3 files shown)


.github/workflows/e2e.yml
```diff
@@ -45,6 +54,9 @@ jobs:
       - name: Start http server
         run: yarn serve:build &
 
+      - name: Set branch and commit as browserstack info
```
mrdrogdrog, line 57: This can also be done in the BrowserStack env action by setting
`build-name`

mrdrogdrog (changes requested):

.github/workflows/e2e.yml
ErikMichelson: Several tests showed that this GH-action field is not working as expected.

Theoretically there are two pre-defined constants "BUILD_INFO" and "REPO_NAME" which could be used for the action properties _build-name_ and _repo-name_. While the action outputs the correct values (BUILD_INFO looks pretty nice btw), they aren't used in the web dashboard. Instead the values of the browserstack.json were used. I removed them but that didn't help either.
The dashboard just says "cypress build" and "untitled cypress project".
I then thought that maybe the constants are a problem and used my manual commit and branch extraction. Result: Cypress build, untitles cypress project.

It seems that the GitHub action for setting up the test environment is just not properly working. It always shows the correct build-name and repo-name in its log but doesn't transmit them to browserstack.
image


For now I'm reverting to jq-ing the browserstack.json - this at least is working.

DerMolly (approved):

mrdrogdrog (approved):

.github/workflows/e2e.yml
mrdrogdrog: sad.. but okay.

PR #938, opened 2021-01-09
fixed copyright template for idea

DerMolly:
### Component/Part
copyright template idea

### Description
This PR fixes the template to always use the current year.

### Steps

- [x] I read the [contribution documentation](https://github.com/hedgedoc/hedgedoc/blob/develop/CONTRIBUTING.md) and
  signed-off my commits to accept the DCO.

### DIFF
--- a/.idea/copyright/hedgedoc.xml
+++ b/.idea/copyright/hedgedoc.xml
@@ -1,6 +1,6 @@
 <component name="CopyrightManager">
   <copyright>
-    <option name="notice" value="SPDX-FileCopyrightText: $today.year The HedgeDoc developers (see AUTHORS file)&#10;&#10;&#83;PDX-License-Identifier: AGPL-3.0-only" />
+    <option name="notice" value="SPDX-FileCopyrightText: ${today.year} The HedgeDoc developers (see AUTHORS file)&#10;&#10;&#83;PDX-License-Identifier: AGPL-3.0-only" />
     <option name="myName" value="hedgedoc" />
   </copyright>
 </component>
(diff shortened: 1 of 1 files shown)


mrdrogdrog (changes requested):
According to https://stackoverflow.com/questions/30618147/how-do-you-insert-the-current-year-into-intellij-idea-copyright-template/30618447
it should be

`${today.year}`

DerMolly:
I tested it and this works, but if you want I can change this.

mrdrogdrog:
Oh it does? Okay, good, then you can see it as suggestion. But tbh I would still prefer to capsule it, so it's clearer to recognize it as template stuff.

(Funfact: It's also good practice in bash to prevent misinterpretation by the runtime)

mrdrogdrog (approved):

.idea/copyright/hedgedoc.xml
```diff
@@ -1,6 +1,6 @@
 <component name="CopyrightManager">
   <copyright>
-    <option name="notice" value="SPDX-FileCopyrightText: 2021 The HedgeDoc developers (see AUTHORS file)&#10;&#10;&#83;PDX-License-Identifier: AGPL-3.0-only" />
+    <option name="notice" value="SPDX-FileCopyrightText: $today.year The HedgeDoc developers (see AUTHORS file)&#10;&#10;&#83;PDX-License-Identifier: AGPL-3.0-only" />
```
ErikMichelson, line 3: I agree with @mrdrogdrog, that the brackets would look more clear here.

DerMolly: done

ErikMichelson (approved):

Issue #815, opened 2020-12-03
KaTeX in first heading breaks TOC and document title


ErikMichelson:
**Description**
KaTeX that is placed into the title of a note doesn't get rendered correctly in the TOC and will cause weird displaying in the browser window title.

**To Reproduce**
1. Visit the current test-instance at https://ui-test.hedgedoc.net/n/features
2. Replace the content with the following one:
```
# $\alpha$-foo
[TOC]
```
3. Have a look at the generated output and the window title

**Expected behavior**
With the example from above, the alpha-character should be displayed in both TOC and window title.

**Screenshots**
![image](https://user-images.githubusercontent.com/52606896/101023725-78417900-3573-11eb-8821-82c26758c3f0.png)

**Additional context**
The problem seems to arise from the circumstance that KaTeX inserts a MathML node into the DOM tree as well as the desired character. Our title-extraction logic could be extended to ignore the math-Tag and thus solve the window title problem. The TOC-problem on the other hand would still reside as the TOC searches for text-tokens in the MarkdownIt-parsing output while KaTeX doesn't output text-tokens. The only option here would be to move the TOC-replacer before the KaTeX-replacer in the order of components. This means that the TOC gets the raw KaTeX code which is at least better than nothing.

DerMolly:
The title extraction is fixed with #946.
But the moving of the toc replacer before the katex-replacer did not work. I think our only way of action is to open an issue with the toc-plugin.

mrdrogdrog:
I'm not sure if this would be necessary. The markdown-it-plugin has a "format" callback, that determines how the link in the toc should be rendered. Eventually we can provide a function that handles the katex symbols correctly.

PR #946, opened 2021-01-11
fixed title extraction to exclude latex code, but include rendered latex

DerMolly:
### Component/Part
title extraction

### Description
This PR improves the title extraction to leave out latex code, but include latex output.
e.g
```
# $\alpha$-foo
```
puts `α-foo` as the title

### Steps

- [x] Added implementation
- [x] Added / updated tests

### Related Issue(s)
concerns #815

### DIFF
--- a/cypress/integration/documentTitle.spec.ts
+++ b/cypress/integration/documentTitle.spec.ts
@@ -101,5 +101,13 @@ describe('Document Title', () => {
       cy.title()
         .should('eq', `${title} asd - HedgeDoc @ ${branding.name}`)
     })
+
+    it('katex code looks right', () => {
+      cy.get('a')
+        .get('@codeinput')
+        .fill(`# $\\alpha$-foo`)
+      cy.title()
+        .should('eq', `α-foo - HedgeDoc @ ${branding.name}`)
+    })
   })
 })
--- a/cypress/integration/documentTitle.spec.ts
+++ b/cypress/integration/documentTitle.spec.ts
@@ -95,16 +95,14 @@ describe('Document Title', () => {
     })
 
     it('markdown syntax removed third', () => {
-      cy.get('a')
-        .get('@codeinput')
+      cy.get('@codeinput')
         .fill(`# ${title} _asd_`)
       cy.title()
         .should('eq', `${title} asd - HedgeDoc @ ${branding.name}`)
     })
 
     it('katex code looks right', () => {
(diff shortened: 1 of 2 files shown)


mrdrogdrog (approved):

ErikMichelson (changes requested):
This seems not to be working properly.

To reproduce:
1. Write `#$\alpha$-foo` in exact that order of letter
2. Select the word `alpha` and replace it with `beta`
3. The preview is updated successfully, but the title shows `\bet-foo`

The CI btw has "Loading..." in the title instead of the character.

DerMolly:
> 1. Write `# $\alpha$-foo` in exact that order of letters
> 
> 2. Select the word `alpha` and replace it with `beta`
> 
> 3. The preview is updated successfully, but the title shows `\bet-foo`
> 
> 
> 
> The CI btw has "Loading..." in the title instead of the character.

Yes the test is not optimal currently.
We found out yesterday, that this seems to be a problem with the renderer and how it works. It needs some rerenders to have the latex code translated in the DOM. And I forgot to account for that in the test as it worked locally.

mrdrogdrog:
Let me elaborate:

KaTeX is an external (quite big) library which gets lazy loaded. While the browser fetches the code, a placeholder (`<span>Loading...</span>`) is shown.
The title gets extracted by a `useEffect` that is bound to the markdown source code. Therefore after the first rendering the effect only sees the placeholder (because katex isn't loaded yet)  and extracts the "Loading...". If you THEN change the markdown content again the effect gets triggered and sees the correct katex code.

mrdrogdrog:
It also seems like katex does some async rendering. I guess the component works the same way we use for mermaid. (Defining a container in react and passing this reference to the katex lib). That would explain why there is broken code in the title sometimes. We have a race condition between the effect and the katex lib.
Unfortunately there isn't much we can do about that :/

DerMolly:
Didn't we add a postRender method, we could use? But that also doesn't catch the async rendered stuff, does it?

mrdrogdrog:
The "postRender" is the useEffect. Effects are always triggered after the rendering. And yes, that wouldn't catch the async katex rendering.

mrdrogdrog:
We could watch the HTML-DOM for changes and trigger the effect again. But I don't think, that it's worth it.

ErikMichelson (approved):

PR #948, opened 2021-01-13
Rewrite tests

mrdrogdrog:
### Component/Part
Cypress tests

### Description
This PR improves the test stability

### Steps

- [x] Added / updated tests


DerMolly (changes requested):
looks good so far, only the toolbar needs some rearrangement. Previously this file was sorted so the test occurred in the same sequence as the button in the toolbar interface. I'd very much like this to be kept.

DerMolly (approved):

ErikMichelson (approved):

PR #957, opened 2021-01-16
Add table formatting on paste of detected table

### LINKED ISSUE #941: Paste table text as markdown table

In GoogleSheets/LibreOfficeCalc/MSExcel you can copy table (sections). You can't simply drop them into hedgedoc though.

When pasting text into the editor, hedgedoc checks if it is formatted like a table and offers to re-write it to look like a real markdown table instead of text with TABs.

- Button/Menu-Item "Paste clipboard as table"
- Context-Menu-Item "Try to format as table" when selecting text

**Additional context**

ErikMichelson:
### Component/Part
Markdown editor

### Description
This PR adds support for pasting tables (e.g. from LibreOffice Calc or MS Excel) into HedgeDoc while reformatting them to markdown tables.

### Steps

- [x] Added implementation
- [x] Added / updated tests
- [x] Extended changelog

### Related Issue(s)
Closes #941

### DIFF
--- a/cypress/integration/fileUpload.spec.ts
+++ b/cypress/integration/fileUpload.spec.ts
@@ -57,24 +57,28 @@ describe('File upload', () => {
               getData: (_: string) => ''
           }
         }
-        cy.get('.CodeMirror-scroll').trigger('paste', pasteEvent)
+        cy.get('.CodeMirror-scroll')
+          .trigger('paste', pasteEvent)
         cy.get('.CodeMirror-activeline > .CodeMirror-line > span')
-          .should('have.text', `![](${imageUrl})`)
+          .should('have.text', `![](${ imageUrl })`)
       })
     })
 
     it('via drag and drop', () => {
-      cy.fixture('acme.png').then((image: string) => {
+      cy.fixture('acme.png')
+        .then((image: string) => {
         const dropEvent = {
           dataTransfer: {
             files: [Cypress.Blob.base64StringToBlob(image, 'image/png')],
             effectAllowed: 'uninitialized'
           }
         }
-        cy.get('.CodeMirror-scroll').trigger('dragenter', dropEvent)
-        cy.get('.CodeMirror-scroll').trigger('drop', dropEvent)
+        cy.get('.CodeMirror-scroll')
+          .trigger('dragenter', dropEvent)
+        cy.get('.CodeMirror-scroll')
+          .trigger('drop', dropEvent)
         cy.get('.CodeMirror-activeline > .CodeMirror-line > span')
-          .should('have.text', `![](${imageUrl})`)
+          .should('have.text', `![](${ imageUrl })`)
       })
     })
   })
@@ -103,7 +107,8 @@ describe('File upload', () => {
         getData: (type = 'text') => testText
       }
     }
-    cy.get('.CodeMirror-scroll').trigger('paste', pasteEvent)
+    cy.get('.CodeMirror-scroll')
+      .trigger('paste', pasteEvent)
     cy.get('.CodeMirror-activeline > .CodeMirror-line > span')
       .should('have.text', `${testText}`)
   })
--- a/cypress/integration/fileUpload.spec.ts
+++ b/cypress/integration/fileUpload.spec.ts
@@ -45,7 +45,7 @@ describe('File upload', () => {
       cy.get('div.btn-group > input[type=file]')
         .attachFile({ filePath: 'acme.png', mimeType: 'image/png' })
       cy.get('.CodeMirror-activeline > .CodeMirror-line > span')
-        .should('have.text', `![](${imageUrl})`)
+        .should('have.text', `![](${ imageUrl })`)
     })
 
     it('via paste', () => {
@@ -110,6 +110,6 @@ describe('File upload', () => {
     cy.get('.CodeMirror-scroll')
       .trigger('paste', pasteEvent)
     cy.get('.CodeMirror-activeline > .CodeMirror-line > span')
-      .should('have.text', `${testText}`)
+      .should('have.text', `${ testText }`)
   })
 })
--- a/cypress/integration/fileUpload.spec.ts
+++ b/cypress/integration/fileUpload.spec.ts
@@ -45,7 +45,7 @@ describe('File upload', () => {
       cy.get('div.btn-group > input[type=file]')
         .attachFile({ filePath: 'acme.png', mimeType: 'image/png' })
       cy.get('.CodeMirror-activeline > .CodeMirror-line > span')
-        .should('have.text', `![](${ imageUrl })`)
+        .should('have.text', `![](${imageUrl})`)
     })
 
     it('via paste', () => {
@@ -53,32 +53,29 @@ describe('File upload', () => {
         .then((image: string) => {
           const pasteEvent = {
             clipboardData: {
-              files: [Cypress.Blob.base64StringToBlob(image, 'image/png')]
-            }
+              files: [Cypress.Blob.base64StringToBlob(image, 'image/png')],
+            getData: (_: string) => ''
           }
-          cy.get('.CodeMirror-scroll')
-            .trigger('paste', pasteEvent)
-          cy.get('.CodeMirror-activeline > .CodeMirror-line > span')
-            .should('have.text', `![](${ imageUrl })`)
-        })
+        }
+        cy.get('.CodeMirror-scroll').trigger('paste', pasteEvent)
+        cy.get('.CodeMirror-activeline > .CodeMirror-line > span')
+          .should('have.text', `![](${imageUrl})`)
+      })
     })
 
     it('via drag and drop', () => {
-      cy.fixture('acme.png')
-        .then((image: string) => {
-          const dropEvent = {
-            dataTransfer: {
-              files: [Cypress.Blob.base64StringToBlob(image, 'image/png')],
-              effectAllowed: 'uninitialized'
-            }
+      cy.fixture('acme.png').then((image: string) => {
+        const dropEvent = {
+          dataTransfer: {
+            files: [Cypress.Blob.base64StringToBlob(image, 'image/png')],
+            effectAllowed: 'uninitialized'
           }
-          cy.get('.CodeMirror-scroll')
-            .trigger('dragenter', dropEvent)
-          cy.get('.CodeMirror-scroll')
-            .trigger('drop', dropEvent)
-          cy.get('.CodeMirror-activeline > .CodeMirror-line > span')
-            .should('have.text', `![](${ imageUrl })`)
-        })
+        }
+        cy.get('.CodeMirror-scroll').trigger('dragenter', dropEvent)
+        cy.get('.CodeMirror-scroll').trigger('drop', dropEvent)
+        cy.get('.CodeMirror-activeline > .CodeMirror-line > span')
+          .should('have.text', `![](${imageUrl})`)
+      })
     })
   })
 
@@ -91,11 +88,10 @@ describe('File upload', () => {
     })
     cy.get('.fa-upload')
       .click()
-    cy.fixture('acme.png')
-      .then(() => {
-        cy.get('input[type=file]')
-          .attachFile({ filePath: 'acme.png', mimeType: 'image/png' })
-      })
+    cy.fixture('acme.png').then(() => {
+      cy.get('input[type=file]')
+        .attachFile({ filePath: 'acme.png', mimeType: 'image/png' })
+    })
     cy.get('.CodeMirror-activeline > .CodeMirror-line > span > span')
       .should('have.text', String.fromCharCode(8203)) //thanks codemirror....
   })
@@ -107,9 +103,8 @@ describe('File upload', () => {
         getData: (type = 'text') => testText
       }
     }
-    cy.get('.CodeMirror-scroll')
-      .trigger('paste', pasteEvent)
+    cy.get('.CodeMirror-scroll').trigger('paste', pasteEvent)
     cy.get('.CodeMirror-activeline > .CodeMirror-line > span')
-      .should('have.text', `${ testText }`)
+      .should('have.text', `${testText}`)
   })
 })
--- a/cypress/integration/fileUpload.spec.ts
+++ b/cypress/integration/fileUpload.spec.ts
@@ -55,31 +55,31 @@ describe('File upload', () => {
             clipboardData: {
               files: [Cypress.Blob.base64StringToBlob(image, 'image/png')],
               getData: (_: string) => ''
+            }
           }
-        }
-        cy.get('.CodeMirror-scroll')
-          .trigger('paste', pasteEvent)
-        cy.get('.CodeMirror-activeline > .CodeMirror-line > span')
-          .should('have.text', `![](${ imageUrl })`)
-      })
+          cy.get('.CodeMirror-scroll')
+            .trigger('paste', pasteEvent)
+          cy.get('.CodeMirror-activeline > .CodeMirror-line > span')
+            .should('have.text', `![](${ imageUrl })`)
+        })
     })
 
     it('via drag and drop', () => {
       cy.fixture('acme.png')
         .then((image: string) => {
-        const dropEvent = {
-          dataTransfer: {
-            files: [Cypress.Blob.base64StringToBlob(image, 'image/png')],
-            effectAllowed: 'uninitialized'
+          const dropEvent = {
+            dataTransfer: {
+              files: [Cypress.Blob.base64StringToBlob(image, 'image/png')],
+              effectAllowed: 'uninitialized'
+            }
           }
-        }
-        cy.get('.CodeMirror-scroll')
-          .trigger('dragenter', dropEvent)
-        cy.get('.CodeMirror-scroll')
-          .trigger('drop', dropEvent)
-        cy.get('.CodeMirror-activeline > .CodeMirror-line > span')
-          .should('have.text', `![](${ imageUrl })`)
-      })
+          cy.get('.CodeMirror-scroll')
+            .trigger('dragenter', dropEvent)
+          cy.get('.CodeMirror-scroll')
+            .trigger('drop', dropEvent)
+          cy.get('.CodeMirror-activeline > .CodeMirror-line > span')
+            .should('have.text', `![](${ imageUrl })`)
+        })
     })
   })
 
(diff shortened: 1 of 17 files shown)


src/components/editor/editor-pane/editor-pane.tsx
```diff
@@ -62,15 +63,31 @@ const onChange = (editor: Editor) => {
...
 
 const onPaste = (pasteEditor: Editor, event: PasteEvent) => {
-  if (event && event.clipboardData && event.clipboardData.files && event.clipboardData.files.length > 0) {
+  if (!event || !event.clipboardData) {
+    return
+  }
+
+  // Check for pasted tables
+  const pasteText = event.clipboardData.getData('text')
+  if (pasteText && pasteText.includes('\t') && pasteText.includes('\n')) {
```
mrdrogdrog, line 83: This is a bit unsafe because these conditions could also apply to python code.
How about an "insert as table" action in the toolbar instead?

DerMolly: We could also check if the code is pasted in between three backticks to prevent python code pasteing…

Also could we check that the code has the same number of tabs in each line?

mrdrogdrog: sry but if we don't have a rock solid proof detection AND a possibility to say "insert as plain text" or "no, that's not what i meant! format as plain text instead (like in word. i love this feature)", i'm against a auto detection of stuff.

DerMolly (approved):

DerMolly (approved):

DerMolly:
Maybe a toast would be good, to tell the user that you autoformated their stuff as a table and how to deactivate that?

mrdrogdrog:
> Maybe a toast would be good, to tell the user that you autoformated their stuff as a table and how to deactivate that?

Did someone say "notifications"? 🌚

src/components/editor-page/editor-pane/editor-pane.tsx
```diff
@@ -103,6 +98,29 @@ export const EditorPane: React.FC<EditorPaneProps & ScrollProps> = ({ onContentC
...
+    }
+
+    // Upload pasted files
+    if (event.clipboardData.files && event.clipboardData.files.length > 0) {
+      event.preventDefault()
+      const files: FileList = event.clipboardData.files
+      if (files && files.length >= 1) {
+        handleUpload(files[0], pasteEditor)
+      }
+    }
```
mrdrogdrog, line 121: Instead of commenting I would extract these code blocks into separate functions

src/components/editor-page/editor-pane/table-extractor.ts
```diff
@@ -0,0 +1,58 @@
...
+  // Tab-indented text should not be matched as a table
+  if (lines.every(line => line.startsWith('\t'))) {
+    return false
+  }
+  // Every line should have the same amount of tabs (table columns)
+  const tabsPerLines = lines.map(line => line.match(/\t/g)?.length ?? 0)
+  return tabsPerLines.every(line => line === tabsPerLines[0])
+}
+
+export const extractTable = (pasteData: string, pasteEditor: Editor): void => {
```
mrdrogdrog, line 33: This method is not only extracting the table, it also insert it into the editor. The name should reflect this. BUT, if you follow clean code, then you should know, that an "and" in the function-name is an indicator for "too much stuff is happening here". So you shouldn't insert the content into the editor here. This also increases the testability.

src/components/editor-page/editor-pane/editor-pane.tsx
```diff
@@ -103,6 +98,29 @@ export const EditorPane: React.FC<EditorPaneProps & ScrollProps> = ({ onContentC
...
+      return
+    }
+
+    // Check for pasted tables
+    const pasteText = event.clipboardData.getData('text')
+    if (smartPasteEnabled && pasteText && isTable(pasteText)) {
+      event.preventDefault()
+      extractTable(pasteText, pasteEditor)
+      return
+    }
```
mrdrogdrog, line 112: https://user-images.githubusercontent.com/6124140/107889213-4a809500-6f11-11eb-905f-c5f350f87e5c.mp4

src/components/editor-page/editor-pane/table-extractor.test.ts
```diff
@@ -0,0 +1,64 @@
+/*
+ * SPDX-FileCopyrightText: 2021 The HedgeDoc developers (see AUTHORS file)
+ *
+ * SPDX-License-Identifier: AGPL-3.0-only
+ */
+
+import { isTable } from './table-extractor'
+
+describe('isTable detection', () => {
```
mrdrogdrog, line 9: Where are the tests for `extractTable` ?

cypress/integration/fileUpload.spec.ts
```diff
@@ -45,40 +45,37 @@ describe('File upload', () => {
...
-          cy.get('.CodeMirror-scroll')
-            .trigger('drop', dropEvent)
-          cy.get('.CodeMirror-activeline > .CodeMirror-line > span')
-            .should('have.text', `![](${ imageUrl })`)
-        })
+        }
+        cy.get('.CodeMirror-scroll').trigger('dragenter', dropEvent)
+        cy.get('.CodeMirror-scroll').trigger('drop', dropEvent)
+        cy.get('.CodeMirror-activeline > .CodeMirror-line > span')
+          .should('have.text', `![](${imageUrl})`)
```
mrdrogdrog, line 77: This seems to be malformatted, because these are chained function calls.

cypress/integration/fileUpload.spec.ts
```diff
@@ -45,40 +45,37 @@ describe('File upload', () => {
       cy.get('div.btn-group > input[type=file]')
         .attachFile({ filePath: 'acme.png', mimeType: 'image/png' })
       cy.get('.CodeMirror-activeline > .CodeMirror-line > span')
-        .should('have.text', `![](${ imageUrl })`)
+        .should('have.text', `![](${imageUrl})`)
```
mrdrogdrog, line 48: This seems to be malformatted, because according to our new editor config templates should have white spaces around the content.

mrdrogdrog (changes requested):

mrdrogdrog:
Sry but I'm still not fully convinced regarding the usability of this feature as long as I can't say "No! I don't want this text to be autoformated. Revert that!"

src/components/editor-page/editor-pane/editor-pane.tsx
ErikMichelson: What's your idea how to detect whether the cursor is inside a code-fence or not?

cypress/integration/fileUpload.spec.ts
ErikMichelson: Hm, I thought I ran the "reformat code" method.

Fixed.

cypress/integration/fileUpload.spec.ts
ErikMichelson: Changed

ErikMichelson:
You might have seen that there is the option to disable this kind of paste-reformatting. But you still want a "revert that" button. Am I right?

If so, we'd need to ensure to replace only the reformatted table (which could have been changed already by someone else) in the document with the plain text and we'd need to store the original paste content somewhere. Sounds not quite trivial.

src/components/editor-page/editor-pane/editor-pane.tsx
mrdrogdrog: Check the line above and below (except blank lines)?

mrdrogdrog:
Tbh.. yes. This is what I would want, because it could confuse the user if they don't know how to disable the feature and if that what they insert is not what they expect. For now this is okay, but imho this would be an excellent candidate for the notifications.

jfowl:
I know you did a lot of work on this already, but what about a button/shortcut which attempts to transform the selected text into a table. sou you would paste your table, select it and then press a button to get a table. Could also be used to create table from CSV code blocks. And it would be explicit.

mrdrogdrog:
I'm not sure if this would be more or less confusing for users who copied a table and expect to insert a table.

mrdrogdrog:
My problem is, that you can't easily revert the paste if the algorithm false-positive detected a table.

jfowl:
What about a modal popping up when pasting to ask you?

mrdrogdrog:
That would interrupt your workflow. Modals which pop up out of nowhere are bad ux.

mrdrogdrog:
Like I said: For now this is .. okay. But we should test test test this. As you can see, unhandled cases (like the paste in the code block) can happen every time.

I still think this would be a great candidate for our notification/toast-idea

src/components/editor-page/editor-pane/editor-pane.tsx
ErikMichelson: Done

src/components/editor-page/editor-pane/table-extractor.ts
ErikMichelson: Changed

src/components/editor-page/editor-pane/editor-pane.tsx
ErikMichelson: I now did the following: Check if either the cursor's line, the one above or the one below have the `m-markdown comment` CodeMirro token, which is applied to codefence blocks.
I need to check the line above and below too as empty lines - although being inside the fence - have no CodeMirror token.

From this results the following:
 1. Reformatting won't trigger if the line above or below starts with `` ``` `` or is inside the fence and has any content.
 2. Reformatting will trigger if you're outside the codefence but directly above the fence-start (next line begins with `` ``` ``).
 3. Same as 2. when you're outside a codefence but the previous line is the end of it.

src/components/editor-page/editor-pane/table-extractor.test.ts
ErikMichelson: Added

src/components/editor-page/editor-pane/table-extractor.ts
```diff
@@ -0,0 +1,57 @@
...
+                             .filter(row => row.trim() !== '')
+  const tableCells = tableRows.reduce((cellsInRow, row, index) => {
+    cellsInRow[index] = row.split('\t')
+    return cellsInRow
+  }, [] as string[][])
+  const arrayMaxRows = createNumberRangeArray(tableCells.length)
+  const arrayMaxColumns = createNumberRangeArray(Math.max(...tableCells.map(row => row.length)))
+
+  const headRow1 = arrayMaxColumns
+    .map(col => col + 1)
```
mrdrogdrog: :banana: I think you can spare this step and use the incremented value in the next step

src/components/editor-page/editor-pane/table-extractor.ts
```diff
@@ -0,0 +1,57 @@
...
+    .map(col => `| #${ col } `)
+    .join('') + '|'
+  const headRow2 = arrayMaxColumns
+    .map(col => col + 1)
```
mrdrogdrog: :banana: same here

src/components/editor-page/editor-pane/table-extractor.ts
```diff
@@ -0,0 +1,57 @@
...
+    cellsInRow[index] = row.split('\t')
+    return cellsInRow
+  }, [] as string[][])
+  const arrayMaxRows = createNumberRangeArray(tableCells.length)
+  const arrayMaxColumns = createNumberRangeArray(Math.max(...tableCells.map(row => row.length)))
+
+  const headRow1 = arrayMaxColumns
+    .map(col => col + 1)
+    .map(col => `| #${ col } `)
+    .join('') + '|'
```
mrdrogdrog, line 43: Why don't you join the column texts with " | " ?

src/components/editor-page/editor-pane/table-extractor.test.ts
```diff
@@ -0,0 +1,90 @@
...
+    const input = '1\t1\n\t0\n\t0\n4\t16\n5\t25\n'
+    expect(isTable(input))
+      .toBe(true)
+  })
+
+  it('valid table with some last cells missing', () => {
+    const input = '1\t1\n2\t\n3\t\n4\t16\n'
+    expect(isTable(input))
+      .toBe(true)
+  })
```
mrdrogdrog, line 63: These case names are different from the others. I'm missing the expectation.

src/components/editor-page/editor-pane/table-extractor.test.ts
```diff
@@ -0,0 +1,90 @@
...
+})
+
+describe('Conversion from "excel" to markdown table', () => {
```
mrdrogdrog, line 66: How about "Tab and newline separated" instead of "excel"?

src/components/editor-page/editor-pane/table-extractor.test.ts
```diff
@@ -0,0 +1,90 @@
...
+    const input = '1\t1\n\t0\n\t0\n4\t16\n'
+    expect(excelTableToMarkdown(input))
+      .toEqual('| #1 | #2 |\n| -- | -- |\n| 1 | 1 |\n|  | 0 |\n|  | 0 |\n| 4 | 16 |')
+  })
+
+  it('works with table with some last cells missing', () => {
+    const input = '1\t1\n2\t\n3\t\n4\t16\n'
+    expect(excelTableToMarkdown(input))
+      .toEqual('| #1 | #2 |\n| -- | -- |\n| 1 | 1 |\n| 2 |  |\n| 3 |  |\n| 4 | 16 |')
+  })
```
mrdrogdrog, line 89: You're only testing positive cases. Aren't there any edge cases?

mrdrogdrog (changes requested):

src/components/editor-page/editor-pane/table-extractor.ts
ErikMichelson: Done

src/components/editor-page/editor-pane/table-extractor.ts
ErikMichelson: Done

src/components/editor-page/editor-pane/table-extractor.test.ts
ErikMichelson: I thought about names and they were all so long.... :D

src/components/editor-page/editor-pane/table-extractor.test.ts
ErikMichelson: What I'll do: Remove the _"valid"_ at the beginning and append _"is valid"_ at the end.

src/components/editor-page/editor-pane/table-extractor.test.ts
mrdrogdrog: Yes but "Excel" is a bit to vague.

mrdrogdrog (changes requested):

src/components/editor-page/editor-pane/table-extractor.ts
ErikMichelson: What would be the benefit? It only requires an additional `|` at the beginning.

mrdrogdrog: You have one string formating step (the template in the map) less?

src/components/editor-page/editor-pane/table-extractor.test.ts
ErikMichelson: To be honest: I can't think of one. Maybe I'm just too uncreative - do you see any edge case I should add?

src/components/editor-page/editor-pane/table-extractor.test.ts
ErikMichelson: Changed

src/components/editor-page/editor-pane/table-extractor.ts
ErikMichelson: Changed when applicable

ErikMichelson: This change only makes things harder. I forgot the `#` in the template literal.
I'll revert my change back to the version you reviewed.

src/components/editor-page/editor-pane/table-extractor.ts
```diff
@@ -0,0 +1,55 @@
...
+  // Tab-indented text should not be matched as a table
+  if (lines.every(line => line.startsWith('\t'))) {
+    return false
+  }
+  // Every line should have the same amount of tabs (table columns)
+  const tabsPerLines = lines.map(line => line.match(/\t/g)?.length ?? 0)
+  return tabsPerLines.every(line => line === tabsPerLines[0])
+}
+
+export const clipboardTableToMarkdown = (pasteData: string): string => {
```
mrdrogdrog, line 31: Clipboard is a weird verb :yum:

src/components/editor-page/editor-pane/table-extractor.test.ts
```diff
@@ -0,0 +1,90 @@
...
+
+  it('works with normal table with newline at end', () => {
+    const input = '1\t1\n2\t4\n3\t9\n4\t16\n'
+    expect(clipboardTableToMarkdown(input))
+      .toEqual('| #1 | #2 |\n| -- | -- |\n| 1 | 1 |\n| 2 | 4 |\n| 3 | 9 |\n| 4 | 16 |')
+  })
+
+  it('works with table with some first cells missing', () => {
+    const input = '1\t1\n\t0\n\t0\n4\t16\n'
```
mrdrogdrog, line 80: These two blocks have different naming pattern. Can we please agree on a pattern? :sweat_smile:

src/components/editor-page/editor-pane/table-extractor.ts
mrdrogdrog: :shrug:

src/components/editor-page/editor-pane/table-extractor.test.ts
mrdrogdrog: What about empty string?

src/components/editor-page/editor-pane/table-extractor.ts
DerMolly: May I suggest `convertCclipboardTableToMarkdown`

mrdrogdrog: nope. to many carbon copies.

mrdrogdrog: `cC`

ErikMichelson: Changed

src/components/editor-page/editor-pane/table-extractor.test.ts
ErikMichelson: Changed

src/components/editor-page/editor-pane/table-extractor.test.ts
ErikMichelson: An empty string won't ever be passed to this method. The method is called only when `isTable(...)` returns true. But that won't return true for an empty string.

Nevertheless, because I know you'd insist on this, I've added a relevant case. :D

mrdrogdrog (changes requested):
You're still not ignoring empty lines in code blocks.

mrdrogdrog (approved):

PR #997, opened 2021-01-30
Use currentColor in SVGs

mrdrogdrog:
### Component/Part
The sync scroll button

### Description
This PR replaces the hard coded colors in the SVGs for the sync scroll buttons with `currentColor`. If you include a SVG directly into the DOM (what we do here), then currentColor is the current text color. This is way more convenient than changing the SVG file.
This doesn't work if you load the SVG with an IMG-Tag.

### Steps

- [x] Added implementation


DerMolly:
Do the SVGs get minified by WebPack? If not we introduce extra bytes of text here that the user needs to load with no benefit…

mrdrogdrog:
The svgs will be inserted directly into the dom. Otherwise this change wouldn't make any sense.

mrdrogdrog:
They are processed as react components.

DerMolly (approved):

ErikMichelson (approved):

PR #1000, opened 2021-01-30
Remove PDF embedding

### LINKED ISSUE #959: Remove PDF-Embedding
I propose to remove the pdf embedding.

Reason 1: 
While running the ui-test on my android tablet I found out that the pdf embedding doesn't work. It seems like chrome for android and firefox for android don't have a pdf plugin, which is needed for the embedding. Safari is also very broken. We want to make the new Frontend mobile friendly and this feature won't work.

Reason 2:
The browsers won't allow embeddings without the right http header anymore soon. See https://HedgeDoc.org/faq#why-cant-i-embed-some-pdfs .

Alternatives:
Replace the embedding with pdf.js rendering, but this solution wouldn't last very long. The content would be fetched asynchronously and could be blocked because of CORS.

mrdrogdrog:
### Component/Part
Markdown renderer

### Description
This PR removes the PDF embedding.

### Steps

- [x] Added implementation
- [x] Extended changelog

### Related Issue(s)
Fixes #959

### Additional Information
:partying_face:  PR 1000!


src/components/markdown-renderer/regex-plugins/replace-legacy-pdf-short-code.ts
```diff
@@ -0,0 +1,20 @@
...
+import MarkdownIt from 'markdown-it/lib'
+
+const finalRegex = /^{%pdf (.*) ?%}$/
+
+export const legacyPdfShortCode: MarkdownIt.PluginSimple = (markdownIt) => {
+  markdownItRegex(markdownIt, {
+    name: 'legacy-pdf-short-code',
+    regex: finalRegex,
+    replace: (match: string) => {
+      return `<a target="_blank" rel="noopener noreferrer" href="${match}">${match}</a>`
```
DerMolly, line 17: Don't we have an external Link component for this?

DerMolly (changes requested):

src/components/markdown-renderer/regex-plugins/replace-legacy-pdf-short-code.ts
mrdrogdrog: This is a markdown it plugin, not a replacer. So I don't know what you mean.

DerMolly: You're right, my bad.

DerMolly (approved):

CHANGELOG.md
```diff
@@ -69,7 +76,7 @@ SPDX-License-Identifier: CC-BY-SA-4.0
 - The sign-in/sign-up functions are now on a separate page
 - The email sign-in/registration does not require an email address anymore but uses a username
 - The history shows both the entries saved in LocalStorage and the entries saved on the server together
-- The gist and pdf embeddings now use a one-click aproach similar to vimeo and youtube
+- The gist embedding now use a one-click approach similar to vimeo and youtube
```
ErikMichelson, line 79:
```suggestion
- The gist embedding now uses a one-click approach similar to vimeo and youtube
```

ErikMichelson (approved):

ErikMichelson (approved):

PR #1020, opened 2021-02-07
Add quote extra plugin

### LINKED ISSUE #648: Allow links in [name=...] syntax
**Description**
This frontend does not allow links inside the name-tag. As 1.6 supports just this, we should fix that.
The minimal example below will not be replaced.

**To Reproduce**
Minimal document:
```markdown
> [name=[Example](//example.com)]
```

**Expected behavior**
The minimal example above (and all other valid cases) will be replaced properly with the name-indicator.

**Additional context**
The problem is the `]` (required for link label) inside a name-tag.
We use the following regex: `/\[name=([^\]]+)]/`
Version 1.6 uses this regex: `/\[name=(.*?)\]/` (thus also allowing empty name-tags)

mrdrogdrog:
### Component/Part
Markdown it plugin for extra quote tags

### Description
This PR adds an improved markdown it plugin that handles the "name" quote tag correctly.

### Steps

- [x] Added implementation
- [x] Added / updated tests

### Related Issue(s)
Fix #648


src/components/markdown-renderer/markdown-it-plugins/parser-debugger.ts
```diff
@@ -10,7 +10,7 @@ export const MarkdownItParserDebugger: MarkdownIt.PluginSimple = (md: MarkdownIt
   if (process.env.NODE_ENV !== 'production') {
     md.core.ruler.push('test', (state) => {
       console.log(state)
-      return true
+      return false
```
DerMolly, line 13: Why is this necessary?

src/components/markdown-renderer/markdown-it-plugins/quote-extra.ts
```diff
@@ -4,44 +4,100 @@
...
+
+      const tokens: Token[] = []
+      state.md.inline.parse(
+        quoteValue,
+        state.md,
+        state.env,
+        tokens
+      )
+
+      //> [name=ChengHan Wu] [time=Sun, Jun 28, 2015 9:59 PM] [color=#907bf7]
```
DerMolly, line 46: Should this be included here?

src/components/markdown-renderer/markdown-it-plugins/quote-extra.ts
```diff
@@ -4,44 +4,100 @@
...
-  regex: /\[name=([^\]]+)]/,
-  replace: (match) => {
-    // ESLint wants to collapse this tag, but then the tag won't be valid html anymore.
-    // noinspection CheckTagEmptyBody
-    return `<span class="quote-extra"><i class="fa fa-user mx-1"></i> ${ match }</span>`
+export const parseLabel = (line: string, start: number, maxPos: number): number | undefined => {
+  for (let pos = start; pos <= maxPos; pos += 1) {
+    if (line[pos] === '=') {
+      return pos
+    }
```
DerMolly, line 101: Maybe some unit tests would a good idea here?

DerMolly (changes requested):

src/components/markdown-renderer/markdown-it-plugins/parser-debugger.ts
mrdrogdrog: Not for this PR, but I didn't want to create a new one for this.
(Technically this is more correct, because it tells markdown it, that this plugin didn't process/change the content)

src/components/markdown-renderer/markdown-it-plugins/quote-extra.ts
mrdrogdrog: This was my demo content. I have commited it by mistake. Removed

ErikMichelson (commented):
Works fine! ✅

Just waiting for the unit test to review.

src/components/markdown-renderer/markdown-it-plugins/quote-extra.ts
mrdrogdrog: Added

DerMolly (approved):

ErikMichelson (approved):

PR #1025, opened 2021-02-08
Add e2e tests for diagrams and embeddings

### LINKED ISSUE #1022: Add E2E tests for "special" embeddings
The different special embeddings (like YouTube, pdf-to-link, Gist, ...) should get E2E tests to ensure that they stay intact.

The pdf-to-link replacer seems to not work anymore after some dependency updates.

ErikMichelson:
### Component/Part
E2E tests

### Description
This PR adds e2e tests for diagrams and embeddings. As most of the diagrams only yield a `svg` element into the DOM, the tests can ensure that some svg element is visible, but can't ensure that it really belongs to the tested diagram.

### Steps

- [x] Added / updated tests

### Related Issue(s)
Closes #1022

### DIFF
--- a/cypress/integration/diagrams.spec.ts
+++ b/cypress/integration/diagrams.spec.ts
@@ -62,6 +62,7 @@ describe('Diagram codeblock gets rendered: ', () => {
     cy.codemirrorFill('```plantuml\nclass Example\n```')
     cy.getMarkdownBody()
       .find('img')
-      .should('have.attr', 'title', 'uml diagram')
+      // PlantUML uses base64 encoded version of zip-deflated PlantUML code in the request URL.
+      .should('have.attr', 'src', 'http://mock-plantuml.local/svg/SoWkIImgAStDuKhEIImkLd2jICmjo4dbSaZDIm6A0W00')
   })
 })
--- a/cypress/integration/diagrams.spec.ts
+++ b/cypress/integration/diagrams.spec.ts
@@ -4,12 +4,12 @@
  * SPDX-License-Identifier: AGPL-3.0-only
  */
 
-describe('Diagram codeblock gets rendered: ', () => {
+describe('Diagram codeblock ', () => {
   beforeEach(() => {
     cy.visitTestEditor()
   })
 
-  it('markmap', () => {
+  it('renders markmap', () => {
     cy.codemirrorFill('```markmap\n- pro\n- contra\n```')
     cy.getMarkdownBody()
       .find('[data-cy=markmap]')
@@ -17,7 +17,7 @@ describe('Diagram codeblock gets rendered: ', () => {
       .should('be.visible')
   })
 
-  it('vega-lite', () => {
+  it('renders vega-lite', () => {
     cy.codemirrorFill('```vega-lite\n{"$schema":"https://vega.github.io/schema/vega-lite/v4.json","data":{"values":[{"a":"","b":28}]},"mark":"bar","encoding":{"x":{"field":"a"},"y":{"field":"b"}}}\n```')
     cy.getMarkdownBody()
       .find('.vega-embed')
@@ -25,7 +25,7 @@ describe('Diagram codeblock gets rendered: ', () => {
       .should('be.visible')
   })
 
-  it('graphviz', () => {
+  it('renders graphviz', () => {
     cy.codemirrorFill('```graphviz\ngraph {\na -- b\n}\n```')
     cy.getMarkdownBody()
       .find('[data-cy=graphviz]')
@@ -33,7 +33,7 @@ describe('Diagram codeblock gets rendered: ', () => {
       .should('be.visible')
   })
 
-  it('mermaid', () => {
+  it('renders mermaid', () => {
     cy.codemirrorFill('```mermaid\ngraph TD;\n    A-->B;\n```')
     cy.getMarkdownBody()
       .find('.mermaid')
@@ -41,7 +41,7 @@ describe('Diagram codeblock gets rendered: ', () => {
       .should('be.visible')
   })
 
-  it('flowchart', () => {
+  it('renders flowcharts', () => {
     cy.codemirrorFill('```flow\nst=>start: Start\ne=>end: End\nst->e\n```')
     cy.getMarkdownBody()
       .find('[data-cy=flowchart]')
@@ -49,7 +49,7 @@ describe('Diagram codeblock gets rendered: ', () => {
       .should('be.visible')
   })
 
-  it('abc', () => {
+  it('renders abc scores', () => {
     cy.codemirrorFill('```abc\nM:4/4\nK:G\n|:GABc dedB:|\n```')
     cy.getMarkdownBody()
       .find('.abcjs-score')
@@ -57,14 +57,14 @@ describe('Diagram codeblock gets rendered: ', () => {
       .should('be.visible')
   })
 
-  it('csv', () => {
+  it('renders csv as table', () => {
     cy.codemirrorFill('```csv delimiter=; header\na;b;c;d\n1;2;3;4\n```')
     cy.getMarkdownBody()
       .find('.csv-html-table')
       .should('be.visible')
   })
 
-  it('plantuml', () => {
+  it('renders plantuml', () => {
     cy.codemirrorFill('```plantuml\nclass Example\n```')
     cy.getMarkdownBody()
       .find('img')
--- a/cypress/integration/diagrams.spec.ts
+++ b/cypress/integration/diagrams.spec.ts
@@ -1,0 +1,67 @@
+/*
+ * SPDX-FileCopyrightText: 2021 The HedgeDoc developers (see AUTHORS file)
+ *
+ * SPDX-License-Identifier: AGPL-3.0-only
+ */
+
+describe('renders diagram', () => {
+  beforeEach(() => {
+    cy.visitTestEditor()
+  })
+
+  it('markmap', () => {
+    cy.codemirrorFill('```markmap\n- pro\n- contra\n```')
+    cy.getMarkdownBody()
+      .find('svg')
+      .should('be.visible')
+  })
+
+  it('vega-lite', () => {
+    cy.codemirrorFill('```vega-lite\n{"$schema":"https://vega.github.io/schema/vega-lite/v4.json","data":{"values":[{"a":"","b":28}]},"mark":"bar","encoding":{"x":{"field":"a"},"y":{"field":"b"}}}\n```')
+    cy.getMarkdownBody()
+      .find('div.vega-embed canvas')
+      .should('be.visible')
+  })
+
+  it('graphviz', () => {
+    cy.codemirrorFill('```graphviz\ngraph {\na -- b\n}\n```')
+    cy.getMarkdownBody()
+      .find('svg')
+      .should('be.visible')
+  })
+
+  it('mermaid', () => {
+    cy.codemirrorFill('```mermaid\ngraph TD;\n    A-->B;\n```')
+    cy.getMarkdownBody()
+      .find('div.mermaid > svg')
+      .should('be.visible')
+  })
+
+  it('flowchart', () => {
+    cy.codemirrorFill('```flow\nst=>start: Start\ne=>end: End\nst->e\n```')
+    cy.getMarkdownBody()
+      .find('svg')
+      .should('be.visible')
+  })
+
+  it('abc', () => {
+    cy.codemirrorFill('```abc\nM:4/4\nK:G\n|:GABc dedB:|\n```')
+    cy.getMarkdownBody()
+      .find('div.abcjs-score > svg')
+      .should('be.visible')
+  })
+
+  it('csv', () => {
+    cy.codemirrorFill('```csv delimiter=; header\na;b;c;d\n1;2;3;4\n```')
+    cy.getMarkdownBody()
+      .find('table.csv-html-table')
+      .should('be.visible')
+  })
+
+  it('plantuml', () => {
+    cy.codemirrorFill('```plantuml\nclass Example\n```')
+    cy.getMarkdownBody()
+      .find('img')
+      .should('have.attr', 'title', 'uml diagram')
+  })
+})
--- a/cypress/integration/diagrams.spec.ts
+++ b/cypress/integration/diagrams.spec.ts
@@ -4,7 +4,7 @@
  * SPDX-License-Identifier: AGPL-3.0-only
  */
 
-describe('renders diagram', () => {
+describe('Diagram codeblock gets rendered: ', () => {
   beforeEach(() => {
     cy.visitTestEditor()
   })
@@ -12,7 +12,7 @@ describe('renders diagram', () => {
   it('markmap', () => {
     cy.codemirrorFill('```markmap\n- pro\n- contra\n```')
     cy.getMarkdownBody()
-      .find('svg')
+      .find('div[data-cy=markmap] svg')
       .should('be.visible')
   })
 
@@ -26,7 +26,7 @@ describe('renders diagram', () => {
   it('graphviz', () => {
     cy.codemirrorFill('```graphviz\ngraph {\na -- b\n}\n```')
     cy.getMarkdownBody()
-      .find('svg')
+      .find('div[data-cy=graphviz] svg')
       .should('be.visible')
   })
 
@@ -40,7 +40,7 @@ describe('renders diagram', () => {
   it('flowchart', () => {
     cy.codemirrorFill('```flow\nst=>start: Start\ne=>end: End\nst->e\n```')
     cy.getMarkdownBody()
-      .find('svg')
+      .find('div[data-cy=flowchart] svg')
       .should('be.visible')
   })
 
@@ -54,7 +54,7 @@ describe('renders diagram', () => {
   it('csv', () => {
     cy.codemirrorFill('```csv delimiter=; header\na;b;c;d\n1;2;3;4\n```')
     cy.getMarkdownBody()
-      .find('table.csv-html-table')
+      .find('.csv-html-table')
       .should('be.visible')
   })
 
--- a/cypress/integration/diagrams.spec.ts
+++ b/cypress/integration/diagrams.spec.ts
@@ -12,42 +12,48 @@ describe('Diagram codeblock gets rendered: ', () => {
   it('markmap', () => {
     cy.codemirrorFill('```markmap\n- pro\n- contra\n```')
     cy.getMarkdownBody()
-      .find('[data-cy=markmap] svg')
+      .find('[data-cy=markmap]')
+      .children()
       .should('be.visible')
   })
 
   it('vega-lite', () => {
     cy.codemirrorFill('```vega-lite\n{"$schema":"https://vega.github.io/schema/vega-lite/v4.json","data":{"values":[{"a":"","b":28}]},"mark":"bar","encoding":{"x":{"field":"a"},"y":{"field":"b"}}}\n```')
     cy.getMarkdownBody()
-      .find('.vega-embed canvas')
+      .find('.vega-embed')
+      .children()
       .should('be.visible')
   })
 
   it('graphviz', () => {
     cy.codemirrorFill('```graphviz\ngraph {\na -- b\n}\n```')
     cy.getMarkdownBody()
-      .find('[data-cy=graphviz] svg')
+      .find('[data-cy=graphviz]')
+      .children()
       .should('be.visible')
   })
 
   it('mermaid', () => {
     cy.codemirrorFill('```mermaid\ngraph TD;\n    A-->B;\n```')
     cy.getMarkdownBody()
-      .find('.mermaid > svg')
+      .find('.mermaid')
+      .children()
       .should('be.visible')
   })
 
   it('flowchart', () => {
     cy.codemirrorFill('```flow\nst=>start: Start\ne=>end: End\nst->e\n```')
     cy.getMarkdownBody()
-      .find('[data-cy=flowchart] svg')
+      .find('[data-cy=flowchart]')
+      .children()
       .should('be.visible')
   })
 
   it('abc', () => {
     cy.codemirrorFill('```abc\nM:4/4\nK:G\n|:GABc dedB:|\n```')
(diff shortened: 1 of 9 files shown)


DerMolly (approved):

src/components/markdown-renderer/replace-components/youtube/youtube-frame.tsx
```diff
@@ -17,7 +17,7 @@ export const YouTubeFrame: React.FC<YouTubeFrameProps> = ({ id }) => {
                        previewContainerClassName={ 'embed-responsive-item' } hoverIcon={ 'youtube-play' }
                        loadingImageUrl={ `//i.ytimg.com/vi/${ id }/maxresdefault.jpg` }>
```
mrdrogdrog, line 18: Please correct the image url as well.

cypress/support/config.ts
```diff
@@ -45,6 +45,7 @@ beforeEach(() => {
         termsOfUse: 'https://example.com/termsOfUse',
         imprint: 'https://example.com/imprint'
       },
+      plantumlServer: 'http://www.plantuml.com/plantuml',
```
mrdrogdrog, line 48: Please don't use a live plantuml server for the test. It slows down the test runner, requires access to the internet and the test could fail if plantuml is down. How about a `cy.intercept` mock?

cypress/integration/shortcodes.spec.ts
```diff
@@ -17,4 +17,13 @@ describe('Short code', () => {
         .should('have.attr', 'href', 'https://www.w3.org/WAI/ER/tests/xhtml/testfiles/resources/pdf/dummy.pdf')
     })
   })
+
```
mrdrogdrog, line 20: what about the youtube shortcode?

cypress/integration/diagrams.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+ *
+ * SPDX-License-Identifier: AGPL-3.0-only
+ */
+
+describe('renders diagram', () => {
+  beforeEach(() => {
+    cy.visitTestEditor()
+  })
+
+  it('markmap', () => {
```
mrdrogdrog, line 12: "Renders diagram markmap" is not a valid sentence. The describe+it calls should build a valid sentence like "Short code for pdf renders a plain link". That's why it's called "it" e.g. "It renders a plain link".
What about
"Diagram embeddings" + "for markdown renders a svg".

cypress/integration/diagrams.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+    cy.codemirrorFill('```markmap\n- pro\n- contra\n```')
+    cy.getMarkdownBody()
+      .find('svg')
+      .should('be.visible')
+  })
+
+  it('vega-lite', () => {
+    cy.codemirrorFill('```vega-lite\n{"$schema":"https://vega.github.io/schema/vega-lite/v4.json","data":{"values":[{"a":"","b":28}]},"mark":"bar","encoding":{"x":{"field":"a"},"y":{"field":"b"}}}\n```')
+    cy.getMarkdownBody()
+      .find('div.vega-embed canvas')
```
mrdrogdrog, line 22: Is the div or the canvas relevant here? The test shouldn't care about how vega-lite renders the diagram.

cypress/integration/diagrams.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+    cy.codemirrorFill('```graphviz\ngraph {\na -- b\n}\n```')
+    cy.getMarkdownBody()
+      .find('svg')
+      .should('be.visible')
+  })
+
+  it('mermaid', () => {
+    cy.codemirrorFill('```mermaid\ngraph TD;\n    A-->B;\n```')
+    cy.getMarkdownBody()
+      .find('div.mermaid > svg')
```
mrdrogdrog, line 36: div

cypress/integration/diagrams.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+    cy.codemirrorFill('```flow\nst=>start: Start\ne=>end: End\nst->e\n```')
+    cy.getMarkdownBody()
+      .find('svg')
+      .should('be.visible')
+  })
+
+  it('abc', () => {
+    cy.codemirrorFill('```abc\nM:4/4\nK:G\n|:GABc dedB:|\n```')
+    cy.getMarkdownBody()
+      .find('div.abcjs-score > svg')
```
mrdrogdrog, line 50: div

cypress/integration/diagrams.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+      .should('be.visible')
+  })
+
+  it('csv', () => {
+    cy.codemirrorFill('```csv delimiter=; header\na;b;c;d\n1;2;3;4\n```')
+    cy.getMarkdownBody()
+      .find('table.csv-html-table')
```
mrdrogdrog, line 57: is the `table` selector necessary?

cypress/integration/diagrams.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+      .should('be.visible')
+  })
+
+  it('plantuml', () => {
+    cy.codemirrorFill('```plantuml\nclass Example\n```')
+    cy.getMarkdownBody()
+      .find('img')
+      .should('have.attr', 'title', 'uml diagram')
```
mrdrogdrog, line 65: Don't rely on titles. I think the URL is a more secure indicator. (Combined with the annotation below)

cypress/integration/linkEmbedder.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+  })
+
+  it('GitHub Gist', () => {
+    cy.codemirrorFill('https://gist.github.com/schacon/1')
+    cy.getMarkdownBody()
+      .find('span.one-click-embedding.gist-frame')
+      .click()
+    cy.getMarkdownBody()
+      .find('iframe')
+      .should('have.attr', 'title', 'gist schacon/1')
```
mrdrogdrog, line 19: Why do you rely on the title and not on the explicit URL you set above?

cypress/integration/linkEmbedder.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+
+describe('links to embeddable content', () => {
+  beforeEach(() => {
+    cy.visitTestEditor()
+  })
+
+  it('GitHub Gist', () => {
+    cy.codemirrorFill('https://gist.github.com/schacon/1')
+    cy.getMarkdownBody()
+      .find('span.one-click-embedding.gist-frame')
```
mrdrogdrog, line 15: And what if we switch from span to div? That would invalidate this test. The `span` selector is redundant here.

cypress/integration/linkEmbedder.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+      .click()
+    cy.getMarkdownBody()
+      .find('iframe')
+      .should('have.attr', 'title', 'gist schacon/1')
+  })
+
+  it('YouTube', () => {
+    cy.codemirrorFill('https://www.youtube.com/watch?v=YE7VzlLtp-4')
+    cy.getMarkdownBody()
+      .find('span.one-click-embedding.embed-responsive-item')
```
mrdrogdrog, line 25: see above.

mrdrogdrog (changes requested):

cypress/integration/diagrams.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+
+describe('renders diagram', () => {
+  beforeEach(() => {
+    cy.visitTestEditor()
+  })
+
+  it('markmap', () => {
+    cy.codemirrorFill('```markmap\n- pro\n- contra\n```')
+    cy.getMarkdownBody()
+      .find('svg')
```
mrdrogdrog, line 15: Tbh i'm not happy with just searching for any svg. If I switch markmap with e.g. graphviz, then this test would still be happy. Please use an (additional) more specific selector. But don't add a new CSS class. Use `data-cy` like we do in other tests.

cypress/integration/diagrams.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+    cy.codemirrorFill('```vega-lite\n{"$schema":"https://vega.github.io/schema/vega-lite/v4.json","data":{"values":[{"a":"","b":28}]},"mark":"bar","encoding":{"x":{"field":"a"},"y":{"field":"b"}}}\n```')
+    cy.getMarkdownBody()
+      .find('div.vega-embed canvas')
+      .should('be.visible')
+  })
+
+  it('graphviz', () => {
+    cy.codemirrorFill('```graphviz\ngraph {\na -- b\n}\n```')
+    cy.getMarkdownBody()
+      .find('svg')
```
mrdrogdrog, line 29: see above

cypress/integration/diagrams.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+    cy.codemirrorFill('```mermaid\ngraph TD;\n    A-->B;\n```')
+    cy.getMarkdownBody()
+      .find('div.mermaid > svg')
+      .should('be.visible')
+  })
+
+  it('flowchart', () => {
+    cy.codemirrorFill('```flow\nst=>start: Start\ne=>end: End\nst->e\n```')
+    cy.getMarkdownBody()
+      .find('svg')
```
mrdrogdrog, line 43: see above

cypress/integration/linkEmbedder.spec.ts
mrdrogdrog, line 25: Can you add a test for the one-click-embedding?

mrdrogdrog (changes requested):

cypress/integration/diagrams.spec.ts
ErikMichelson: Well, "Short code for pdfs" isn't a sentence as well, just to name one example.
A complete sentence seems a bit verbose. What about the following schema?

Diagram embedding renders correctly: <markmap|vega-lite|...>

cypress/integration/diagrams.spec.ts
ErikMichelson: The div alone does not say that the diagram is rendered. Matching a canvas alone is a bit too vague.

cypress/integration/diagrams.spec.ts
ErikMichelson: Same as above

cypress/integration/diagrams.spec.ts
ErikMichelson: Same as above

cypress/integration/diagrams.spec.ts
ErikMichelson: Maybe not, but isn't it better to fence the test-element stricter than lesser?

cypress/integration/linkEmbedder.spec.ts
ErikMichelson: What do you mean with this?

cypress/integration/linkEmbedder.spec.ts
ErikMichelson: Because the URL is a dynamically generated base64 url of our iframe html code.

cypress/integration/diagrams.spec.ts
DerMolly: I think Tilman is suggesting to use `.mermaid > svg` or just `.mermaid` for the test. I guess if one wants the svg depends on how the error case looks. If in case of an error the output would also contain some HTML Tag with the class `.mermaid`, but not an `svg` this should be included…

cypress/integration/diagrams.spec.ts
mrdrogdrog: Just use the class?

cypress/integration/diagrams.spec.ts
mrdrogdrog: Then you should just use the class or use a data-cy tag. We're testing the replacer, not mermaid.

cypress/integration/diagrams.spec.ts
mrdrogdrog: same as above

cypress/integration/linkEmbedder.spec.ts
mrdrogdrog: Well. fair. But then I would rather use a data-cy attribute. The title can change at any time.

cypress/integration/linkEmbedder.spec.ts
mrdrogdrog: That the one-click-embedding component reacts to clicks, renders a preview image, loads the content behind it, etc.

src/components/markdown-renderer/replace-components/youtube/youtube-frame.tsx
ErikMichelson: Done

cypress/integration/shortcodes.spec.ts
ErikMichelson: Added

cypress/support/config.ts
ErikMichelson: How do you want to mock PlantUML? Returning a predefined "diagram" image url?

cypress/integration/linkEmbedder.spec.ts
ErikMichelson: Done

cypress/integration/linkEmbedder.spec.ts
ErikMichelson: Done

cypress/integration/diagrams.spec.ts
ErikMichelson: `data-cy` added

cypress/integration/diagrams.spec.ts
ErikMichelson: `data-cy` added

cypress/integration/diagrams.spec.ts
ErikMichelson: `data-cy` added

cypress/integration/linkEmbedder.spec.ts
ErikMichelson: If you look a few lines below your annotation, you'll find the click-reaction and content-loading. Just the preview image needs to be tested yet.

cypress/support/config.ts
mrdrogdrog: Yes

ErikMichelson: Done

cypress/integration/diagrams.spec.ts
ErikMichelson: Done

cypress/integration/diagrams.spec.ts
ErikMichelson: Changed

cypress/integration/linkEmbedder.spec.ts
ErikMichelson: Uses data-cy attribute now

cypress/integration/linkEmbedder.spec.ts
ErikMichelson: Tests for preview images added

cypress/integration/diagrams.spec.ts
```diff
@@ -12,42 +12,42 @@ describe('Diagram codeblock gets rendered: ', () => {
...
       .should('be.visible')
   })
 
   it('abc', () => {
     cy.codemirrorFill('```abc\nM:4/4\nK:G\n|:GABc dedB:|\n```')
     cy.getMarkdownBody()
-      .find('div.abcjs-score > svg')
+      .find('.abcjs-score > svg')
       .should('be.visible')
   })
```
mrdrogdrog: I'm still not sure if it's a good idea to expect elements, that we don't control or create. The SVGs and the canvas you're looking for here are not part of our code, but of the 3rd-party rendering libs themself.
I understand, that you want to test if our code triggers the lib code but imho this is useless as long as we can't verify what the user is seeing. Also if e.g. abcjs switches from svg to canvas, then this test will fail and we still don't know if the output is correct.

cypress/integration/linkEmbedder.spec.ts
```diff
@@ -15,24 +15,38 @@ describe('Link gets replaced with embedding: ', () => {
...
+      .should('be.visible')
   })
 
   it('YouTube', () => {
     cy.codemirrorFill('https://www.youtube.com/watch?v=YE7VzlLtp-4')
     cy.getMarkdownBody()
-      .find('.one-click-embedding.embed-responsive-item')
+      .find('.one-click-embedding-preview')
+      .should('have.attr', 'src', 'https://i.ytimg.com/vi/YE7VzlLtp-4/maxresdefault.jpg')
+      .parent()
```
mrdrogdrog, line 27: I think we should find a way to test the click-shield-component itself instead of using a specific implementation. But I don't know how. Maybe there are unit tests for React? :thinking:  Can you add a todo here for this?

mrdrogdrog (changes requested):

cypress/integration/diagrams.spec.ts
mrdrogdrog: Well that's true. Okay, how about this: Describe blocks don't have to be part of the senteces, but should specify the parent describe block. But I insist, that the it-calls build correct senteces. It is a convention in testing.

cypress/integration/diagrams.spec.ts
ErikMichelson: It's true that we don't control the specific elements. But we sometimes control the parent container that has the class name or data-cy attribute and show for example error messages inside it on failures.
When just testing for the parent container, we don't make sure that the library really runs. That's not end-to-end testing but end-to-one-step-before-end.
While it's also true that maybe one day a dependency changes their rendering method, it's actually good that the test fails in that case. It reminds us to check whether the rendering works still after the dependency restructuring. And changing one selector in a cypress test inside a renovate PR isn't a major deal.

cypress/integration/linkEmbedder.spec.ts
ErikMichelson: Sounds almost like [storybook tests](https://storybook.js.org/docs/react/workflows/testing-with-storybook).

I added a TODO entry.

cypress/integration/diagrams.spec.ts
ErikMichelson: It's important to have a good structure.
And a complete sentence might read good in single occasions and when you read it alone standing. But if you have dozens of log lines - like "Component xabc renders as expected", "Component ydfghj renders as expected" - it's much harder to get an overview of the actual component tested, especially as similar words are not aligned beneath each other.

I therefore stay at the position that test labels like "renders as expected: xabc", "renders as expected: ydfghj" fulfill the same purpose while being easier to read.

mrdrogdrog: I'm seeing this the opposite way. For me a senteces if much better readable. Especially in the output.

Sry but then I can't approve this, if we don't agree on a syntax, because then some tests are names like you want and some like I want and maybe @DerMolly also has another opinon.

cypress/integration/diagrams.spec.ts
mrdrogdrog: Tbh actually also doing an "end-to-one-step-before-end" test, because you're not checking if the lib renders the output correctly. To be a complete end to end test, you have to check if e.g. some words are contained in the output.

mrdrogdrog: We don't know if the svg you're checking contains any reasonable content. And as long as we don't check this, there is no point in checking the existence of the svg.

cypress/integration/diagrams.spec.ts
DerMolly: okay what about
describe: `rendering diagram`
	it: `markmap works`

cypress/integration/diagrams.spec.ts
ErikMichelson: Most of the libs probably have their own tests that their rendering output works. Therefore we don't need to test the libs that the use. We need to test whether we trigger the libs properly to render. When looking for a div that _could_ contain a diagram, it doesn't mean that the diagram library has been triggered. It also could be that the replacer was refactored but the lib call itself was forgotten.
When checking for the existence of just some rendering, we at least ensure that the rendering lib has been called.

cypress/integration/diagrams.spec.ts
ErikMichelson: @mrdrogdrog I don't understand why suddenly now you want to enforce your way. If we had agreed to always use such a structure or if every other test follows the same schema, it would be a different thing. But: From 131 cypress tests being in main branch less than the half - more exact 60 - are written as a sentence.

Moreover, I'll show you why I don't agree on sentences. Look at the image below. On a quick glimpse which of the tests could have failed you just see "shows correct line number" and have to read the beginning of your sentence above that.

image

cypress/integration/diagrams.spec.ts
mrdrogdrog: Theeeen how about "div > *" ? :thinking:

cypress/integration/diagrams.spec.ts
mrdrogdrog: Well. It's a pity that only jest seems to follow the convention.

We never agreed on a syntax, therefore I started to use the conventional syntax. This is why I started "now". 
Anyway, we should agree on something.

Imho the test case name should include the expectation not the object that is tested. That's what a describe block is for.

mrdrogdrog: like `"Diagram embeddings" + "renders markmap"`

cypress/integration/diagrams.spec.ts
mrdrogdrog: In cypress it should be
```
      .children()
      .should("exist")
```

ErikMichelson: I did this with the `.children()` method you mentioned.

cypress/integration/diagrams.spec.ts
ErikMichelson: > like `"Diagram embeddings" + "renders markmap"`

Did that.

mrdrogdrog (approved):

PR #1201, opened 2021-04-21
Fix Communication between frontend and backend

DerMolly:
### Component/Part
api

### Description
This PR fixes some small issues with the api that came up while letting backend and frontend work together…

### Steps

- [x] Added implementation

### Related Issue(s)
the corresponding PR in the backend https://github.com/hedgedoc/hedgedoc/pull/1162


public/api/private/config
```diff
@@ -36,12 +36,13 @@
     "imprint": "https://example.com/imprint"
   },
   "version": {
-    "version": "mock",
-    "sourceCodeUrl": "https://www.youtube.com/watch?v=dQw4w9WgXcQ",
-    "issueTrackerUrl": "https://www.youtube.com/watch?v=dQw4w9WgXcQ"
```
mrdrogdrog, line 41: Why did you remove these?

mrdrogdrog (changes requested):

public/api/private/config
DerMolly: The backend route does not provide those and I think the whole thing can be removed (the version modal needs changes for that), because this was built under the assumption that frontend and backend wouldn't be delivered together. As this is now the case this does not make sense anymore…

mrdrogdrog: So we hardcode the urls now?.. or put them into the version json?

DerMolly: I'm not to sure… I'd suggest removing the two columns and just have the version and the link to the code repo and the issue tracker. Those should the hardcoded then…

Maybe it would be a good idea to add a issue template about GUI issues or something similiar in the hedgedoc/hedgedoc repo which link to an issue in the react-client?

mrdrogdrog: Please don't hard code them. Put them at least into a json file. Or read read them from the package.json.

DerMolly: Seems like a reasonable way of doing things…

ErikMichelson: Reminder that we've got a links.json file...

src/api/notes/index.ts
```diff
@@ -23,11 +23,18 @@ export interface Note {
...
-  })
+  if (process.env.NODE_ENV === 'development') {
+    response = await fetch(getApiUrl() + `/notes/${ noteId }-get`, {
+      ...defaultFetchConfig
+    })
+  } else {
+    response = await fetch(getApiUrl() + `/notes/${ noteId }`, {
+      ...defaultFetchConfig
+    })
+  }
```
ErikMichelson, line 37: Couldn't this be shortened by usage of a ternary operator?

DerMolly: Something like this?
```
const response = await fetch(getApiUrl() + `/notes/${ noteId }${process.env.NODE_ENV === 'development' ? '-get' : ''}`, {
      ...defaultFetchConfig
    })
```

public/api/private/config
DerMolly: I reworked this. Have a look

src/api/notes/index.ts
DerMolly: Changed to the mentioned solution

mrdrogdrog: imho this line is much harder to read now. Could you please merge the getApiURL into the string template and extract the suffix into a var?

mrdrogdrog: Also: do you need the spread syntax here?

ErikMichelson: The spread syntax isn't needed if the `{ }` around it get removed.

DerMolly: I think it's best that this is kept as is. We do it in all fetch calls and it makes it easy to later add extra option if we need them.

mrdrogdrog (approved):

mrdrogdrog (changes requested):
Cypress tests is failing because the intercept in cypress/support/visit-test-editor.ts#L20 is returning the wrong object.

mrdrogdrog (approved):

public/api/private/notes/old
```diff
@@ -1,13 +1,18 @@
 {
```
ErikMichelson, line 1: Shouldn't this file be named `old-get`?

src/utils/is-test-mode.ts
```diff
@@ -7,3 +7,7 @@
 export const isTestMode = (): boolean => {
   return !!process.env.REACT_APP_TEST_MODE
 }
+
+export const isMockMode = (): boolean => {
```
ErikMichelson, line 11: The name of the file is a bit misleading as it now doesn't contain the isTestMode exclusively.

ErikMichelson (changes requested):

public/api/private/notes/old
mrdrogdrog: do we even need this file?

ErikMichelson: Well, the intention of this old note was to test whether the redirect from `/old` to `/n/old` works (cypress test missing btw [but not in this PR]).

DerMolly: Well we should be able to test the redirector without a real backend.

src/utils/is-test-mode.ts
DerMolly: renamed to `test-modes`

public/api/private/notes/old
DerMolly: fixed

ErikMichelson:
I just realized we'd need a corresponding PR in hedgedoc/react-client-test-page to not break the CI, as the backend doesn't contain the source code and issue tracker URLs in its config anymore.

src/utils/test-modes.ts
```diff
@@ -0,0 +1,13 @@
+/*
+ * SPDX-FileCopyrightText: 2021 The HedgeDoc developers (see AUTHORS file)
+ *
+ * SPDX-License-Identifier: AGPL-3.0-only
+ */
+
+export const testModes = (): boolean => {
```
ErikMichelson, line 7: This is wrong.

ErikMichelson (approved):

ErikMichelson (approved):

PR #1216, opened 2021-04-25
Fetch banner.txt from public URL instead of config

mrdrogdrog:
### Component/Part
MOTD-Banner

### Description
This PR changes the behaviour for the motd banner. The content is fetched from the a file via ajax instead of reading it from the config.

### Steps

- [x] Added implementation
- [x] Added / updated tests


src/components/application-loader/initializers/configLoader.ts
```diff
@@ -21,14 +23,8 @@ export const loadAllConfig: (baseUrl: string) => Promise<void> = async (baseUrl)
...
-  const banner = config.banner
-  if (banner.text !== '') {
-    const lastAcknowledgedTimestamp = window.localStorage.getItem('bannerTimeStamp') || ''
-    setBanner({
-      ...banner,
-      show: banner.text !== '' && banner.timestamp !== lastAcknowledgedTimestamp
-    })
-  }
+  fetchAndSetBanner(backendUrl)
+    .catch(console.error.bind(this))
```
ErikMichelson, line 27: As this does not belong to the config anymore, why do you include it into the configLoader.ts instead of adding it as an own initializer task?

src/components/common/motd-banner/motd-banner.tsx
```diff
@@ -1,41 +1,56 @@
...
-          onClick={ dismissBanner }>
-          <ForkAwesomeIcon icon='times'/>
-        </Button>
-      </Alert>
-    </ShowIf>
+    <Alert data-cy={ 'motd-banner' } variant="primary" dir="auto"
+           className="mb-0 text-center d-flex flex-row justify-content-center">
+      <Link to="/s/banner" className="flex-grow-1 align-self-center text-black">
+        { bannerState.banner.text }
+      </Link>
```
ErikMichelson, line 45: The link to /s/banner doesn't make sense anymore when the banner content isn't the same thing.
Maybe we want to allow placing a custom URL in the last line of the banner.txt and use that as link?

src/redux/banner/methods.ts
```diff
@@ -10,7 +10,7 @@ import { BannerActionType, BannerState, SetBannerAction } from './types'
 export const setBanner = (state: BannerState): void => {
   const action: SetBannerAction = {
     type: BannerActionType.SET_BANNER,
-    state
+    state: state
```
ErikMichelson, line 13: This shouldn't be needed and is inconsistent with other redux set methods.

cypress/integration/banner.spec.ts
```diff
@@ -4,31 +4,79 @@
...
+    cy.get('[data-cy="motd-banner"]')
+      .contains(bannerMocKContent)
+    cy.get('button[data-cy="motd-dismiss"]')
       .click()
       .then(() => {
-        expect(localStorage.getItem('bannerTimeStamp'))
+        expect(localStorage.getItem('banner.etag'))
           .to
-          .equal(banner.timestamp)
+          .be
```
ErikMichelson, line 49: Why do you add the `.be`?

If you're aiming for "valid sentences", it needs to be "expect banner.etag to be equal to mock_e_tag". This is obviously not possible, therefore the usage of "equal" as a verb sounds better: "expect banner.etag to equal mock_e_tag".

cypress/integration/banner.spec.ts
```diff
@@ -4,31 +4,79 @@
  * SPDX-License-Identifier: AGPL-3.0-only
  */
 
-import { banner } from '../support/config'
+const BANNER_LOCAL_STORAGE_KEY = 'banner.etag'
+const MOCK_E_TAG = 'mockETag'
+const bannerMocKContent = 'This is the mock banner call'
```
ErikMichelson, line 9:
```suggestion
const bannerMockContent = 'This is the mock banner call'
```

cypress/integration/banner.spec.ts
```diff
@@ -4,31 +4,79 @@
...
+
+  it('won\'t show again on reload', () => {
+    cy.get('[data-cy="motd-banner"]')
+      .contains(bannerMocKContent)
+    cy.get('button[data-cy="motd-dismiss"]')
+      .click()
+      .then(() => {
+        expect(localStorage.getItem(BANNER_LOCAL_STORAGE_KEY))
+          .to
+          .be
```
ErikMichelson, line 66: Same thing as above. No `.be`.

cypress/integration/banner.spec.ts
```diff
@@ -4,31 +4,79 @@
...
   beforeEach(() => {
+    cy.intercept({
+      method: 'GET',
+      url: '/public/banner.txt'
+    }, {
+      statusCode: 200,
+      headers: { 'ETag': MOCK_E_TAG },
+      body: bannerMocKContent
+    })
+      .as('getBanner')
```
ErikMichelson, line 21: You don't use this alias.

src/components/application-loader/initializers/configLoader.ts
```diff
@@ -6,13 +6,15 @@
 
 import { getConfig } from '../../../api/config'
 import { setApiUrl } from '../../../redux/api-url/methods'
-import { setBanner } from '../../../redux/banner/methods'
 import { setConfig } from '../../../redux/config/methods'
 import { getAndSetUser } from '../../login-page/auth/utils'
+import { fetchAndSetBanner } from './fetch-and-set-banner'
 
 export const loadAllConfig: (baseUrl: string) => Promise<void> = async (baseUrl) => {
+  const backendUrl = (process.env.REACT_APP_BACKEND || baseUrl)
```
ErikMichelson, line 14: Wouldn't it make more sense to extract this into a common available variable?
(Like the `utils/isTestMode`)

src/redux/banner/types.ts
```diff
@@ -15,11 +15,13 @@ export interface BannerActions extends Action<BannerActionType> {
...
   state: BannerState;
 }
 
 export interface BannerState {
-  show: boolean
-  text: string
-  timestamp: string
+  banner: undefined | {
+    text: string | undefined
+    etag: string | null
```
ErikMichelson, line 25: I don't like the fact that the BannerState object contains a single property "banner" which again contains the wanted values.

ErikMichelson (changes requested):

src/redux/banner/methods.ts
mrdrogdrog: Actually it would be better if we add this to the other actions. So you can't mix up actions with wrong interfaces

cypress/integration/banner.spec.ts
mrdrogdrog: This was already there in the original code.

src/components/application-loader/initializers/configLoader.ts
mrdrogdrog: Tbh I wasn't sure if we need this in more than one places. Most of the time we just need the API url. And this gets set in this method.

src/redux/banner/types.ts
mrdrogdrog: I'm also not very happy with that but i could think of another name. Top level redux states must not be undefined.

src/components/application-loader/initializers/configLoader.ts
mrdrogdrog: There was a reason why I did that.. but i don't know why anymore. I'll fix that.

src/components/common/motd-banner/motd-banner.tsx
mrdrogdrog: Hm. We could use markdown for the banner and could define this in it's frontmatter.

src/redux/banner/types.ts
DerMolly: Then just set text to "" and date to new Date(0) and don't remove the show variable?

src/redux/banner/methods.ts
ErikMichelson: Hm? This is just the ES6 short syntax?

mrdrogdrog: That's typescript. If you define a type in an superior interface and a specific value type in children interfaces then typescript can type interfer the other attributes of the interfaces.

src/redux/banner/types.ts
mrdrogdrog: Not sure yet. I need to differentiate between "not fetched yet" and "fetched but ignored". But yeah. Maybe i should just use an empty string. I mean: who would show a banner with no content?

DerMolly: We explicitly don't show the banner if the text is empty. At least we did so until now…

src/redux/banner/methods.ts
mrdrogdrog: Oh you mean THIS! sry I though you meant something else. Ignore my comment then.

src/components/application-loader/initializers/configLoader.ts
mrdrogdrog: I moved this var

src/components/common/motd-banner/motd-banner.tsx
mrdrogdrog: I removed the Link for now. If we still want to have it then we can readd that in another PR.

DerMolly (approved):

src/components/application-loader/initializers/fetch-and-set-banner.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  if (response.status !== 200) {
+    return
+  }
+
+  const bannerText = await response.text()
+
+  const lastModified = response.headers.get('Last-Modified')
+  if (!lastModified) {
+    console.warn('ETag not found for banner.txt!')
```
ErikMichelson, line 45:
```suggestion
    console.warn('Last-Modified header not found for banner.txt!')
```

ErikMichelson (commented):
Seems to have some rebase issues? Or why does this PR add files that have been deleted in other PRs?

src/components/application-loader/initializers/fetch-and-set-banner.ts
mrdrogdrog: done

ErikMichelson (approved):

DerMolly (approved):

PR #1229, opened 2021-05-02
Add RevealOptions for revealjs 4.1.0 

DerMolly:
### Component/Part
external types

### Description
This PR adds RevealOptions for reveal 4.1.0 

### Steps

- [x] Added implementation

### DIFF
--- a/src/external-types/reveal.js/index.d.ts
+++ b/src/external-types/reveal.js/index.d.ts
@@ -4,9 +4,180 @@
  * SPDX-License-Identifier: AGPL-3.0-only
  */
 
+import {EasingFunction, Property} from 'csstype'
+
 declare module 'reveal.js' {
+  export type AutoAnimateStyles = 'background-color' | 'border-color' | 'border-radius' | 'border-width' | 'color' | 'font-size' | 'letter-spacing' | 'line-height' | 'opacity' | 'outline' | 'outline-offset' | 'padding'
+
+  export type Transition = 'none' | 'fade' | 'slide' | 'convex' | 'concave' | 'zoom'
+
+  export type ControlsBackArrows = 'faded' | 'hidden' | 'visible'
+
+  export type ControlsLayout = 'bottom-right' | 'edges'
+
+  export type NavigationMode = 'default' | 'linear' | 'grid'
+
+  export type ShowSlideNumber = 'all' | 'print' | 'speaker'
+
+  export type SlideNumber = 'h.v' | 'h/v' | 'c' | 'c/t'
+
+  export type TransitionSpeed = 'default' | 'slow' | 'fast'
+
   export interface RevealOptions {
-    plugins?: string[]
+    autoAnimate?: boolean,
+    // Time between auto-animate transitions
+    autoAnimateDuration?: number,
+    // Auto-animate transition function
+    autoAnimateEasing?: EasingFunction,
+    // CSS properties that can be auto-animated. Position & scale
+    // is matched separately so there's no need to include styles
+    // like top/right/bottom/left, width/height or margin.
+    autoAnimateStyles?: AutoAnimateStyles[],
+    // Global override for autolaying embedded media (video/audio/iframe)
+    // - null:   Media will only autoplay if data-autoplay is present
+    // - true:   All media will autoplay, regardless of individual setting
+    // - false:  No media will autoplay, regardless of individual setting
+    autoPlayMedia?: boolean | null,
+    // Controls automatic progression to the next slide
+    // - 0:      Auto-sliding only happens if the data-autoslide HTML attribute
+    //           is present on the current slide or fragment
+    // - 1+:     All slides will progress automatically at the given interval
+    // - false:  No auto-sliding, even if data-autoslide is present
+    autoSlide?: 0 | 1 | false,
+    // Stop auto-sliding after user input
+    autoSlideStoppable?: boolean,
+    // Transition style for full page slide backgrounds
+    backgroundTransition?: Transition,
+    // Vertical centering of slides
+    center?: boolean,
+    // Display presentation control arrows
+    controls?: boolean,
+    // Visibility rule for backwards navigation arrows
+    controlsBackArrows?: ControlsBackArrows,
+    // Determines where controls appear
+    controlsLayout?: ControlsLayout,
+    // Help the user learn the controls by providing hints, for example by
+    // bouncing the down arrow when they first encounter a vertical slide
+    controlsTutorial?: boolean,
+    // Specify the average time in seconds that you think you will spend
+    // presenting each slide. This is used to show a pacing timer in the
+    // speaker view
+    defaultTiming?: number | null,
+    // Disables the default reveal.js slide layout (scaling and centering)
+    // so that you can use custom CSS layout
+    disableLayout?: boolean,
+    // The display mode that will be used to show slides
+    display?: Property.Display,
+    // Flags if the presentation is running in an embedded mode,
+    // i.e. contained within a limited portion of the screen
+    embedded?: boolean,
+    // Focuses body when page changes visibility to ensure keyboard shortcuts work
+    focusBodyOnPageVisibilityChange?: boolean,
+    // Flags whether to include the current fragment in the URL,
+    // so that reloading brings you to the same fragment position
+    fragmentInURL?: boolean, // ToDo: see [1]
+    // Turns fragments on and off globally
+    fragments?: boolean,
+    // Add the current slide number to the URL hash so that reloading the
+    // page/copying the URL will return you to the same slide
+    hash?: boolean, // ToDo: see [1]
+    // Use 1 based indexing for # links to match slide number (default is zero
+    // based)
+    hashOneBasedIndex?: boolean, // ToDo: see [1]
+    // Flags if we should show a help overlay when the question-mark
+    // key is pressed
+    help?: boolean,
+    // Time before the cursor is hidden (in ms)
+    hideCursorTime?: number,
+    // Hide cursor if inactive
+    hideInactiveCursor?: boolean,
+    // Enable keyboard shortcuts for navigation
+    keyboard?: boolean,
+    // Loop the presentation
+    loop?: boolean,
+    // Number of slides away from the current that are visible on mobile
+    // devices. It is advisable to set this to a lower number than
+    // viewDistance in order to save resources
+    mobileViewDistance?: number,
+    // Enable slide navigation via mouse wheel
+    mouseWheel?: boolean,
+    // Changes the behavior of our navigation directions.
+    //
+    // "default"
+    // Left/right arrow keys step between horizontal slides, up/down
+    // arrow keys step between vertical slides. Space key steps through
+    // all slides (both horizontal and vertical).
+    //
+    // "linear"
+    // Removes the up/down arrows. Left/right arrows step through all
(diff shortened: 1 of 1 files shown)


src/external-types/reveal.js/index.d.ts
```diff
@@ -4,9 +4,199 @@
...
+    // Determines where controls appear
+    controlsLayout?: ControlsLayout,
+    // Help the user learn the controls by providing hints, for example by
+    // bouncing the down arrow when they first encounter a vertical slide
+    controlsTutorial?: boolean,
+    // Specify the average time in seconds that you think you will spend
+    // presenting each slide. This is used to show a pacing timer in the
+    // speaker view
+    defaultTiming?: number | null,
+    /* disableLayout?: boolean, // ToDo: I think this is not needed */
```
DerMolly, line 71: > Disables the default reveal.js slide layout (scaling and centering)
> so that you can use custom CSS layout

Do we want to omit CSS completely? I guess with the iframe it would be possible to define ones own theme in HedgeDoc…

src/external-types/reveal.js/index.d.ts
```diff
@@ -4,9 +4,199 @@
...
+    // The display mode that will be used to show slides
+    display?: Property.Display,
+    // Flags if the presentation is running in an embedded mode,
+    // i.e. contained within a limited portion of the screen
+    embedded?: boolean,
+    // Focuses body when page changes visibility to ensure keyboard shortcuts work
+    focusBodyOnPageVisibilityChange?: boolean,
+    // Flags whether to include the current fragment in the URL,
+    // so that reloading brings you to the same fragment position
+    /* fragmentInURL?: boolean, // ToDo: I think this is not needed */
```
DerMolly, line 81: > Flags whether to include the current fragment in the URL,
> so that reloading brings you to the same fragment position

I'm not sure this is to easy to implement. Do we want to include this then?

src/external-types/reveal.js/index.d.ts
```diff
@@ -4,9 +4,199 @@
...
+    // Turns fragments on and off globally
+    fragments?: boolean,
+    // Add the current slide number to the URL hash so that reloading the
+    // page/copying the URL will return you to the same slide
+    /* hash?: boolean, // ToDo: I think this is not needed */
```
DerMolly, line 86: > Add the current slide number to the URL hash so that reloading the
> page/copying the URL will return you to the same slide

I think this also depends on, if we want `fragmentInURL`…

src/external-types/reveal.js/index.d.ts
```diff
@@ -4,9 +4,199 @@
...
+    // Use 1 based indexing for # links to match slide number (default is zero
+    // based)
+    /* hashOneBasedIndex?: boolean, // ToDo: I think this is not needed */
```
DerMolly, line 89: > Use 1 based indexing for # links to match slide number (default is zero based)

I think this also depends on, if we want `fragmentInURL`…

src/external-types/reveal.js/index.d.ts
```diff
@@ -4,9 +4,199 @@
...
+    // Flags if we should show a help overlay when the question-mark
+    // key is pressed
+    help?: boolean,
+    // Time before the cursor is hidden (in ms)
+    hideCursorTime?: number,
+    // Hide cursor if inactive
+    hideInactiveCursor?: boolean,
+    // Push each slide change to the browser history.  Implies `hash: true`
+    /* history?: boolean, // ToDo: I think this is not needed */
```
DerMolly, line 98: > Push each slide change to the browser history. Implies `hash: true`

I think this also depends on, if we want `hash` and therefore `fragmentInURL`…

src/external-types/reveal.js/index.d.ts
```diff
@@ -4,9 +4,199 @@
...
+    // Enable keyboard shortcuts for navigation
+    keyboard?: boolean,
+    // Optional function that blocks keyboard events when retuning false
+    //
+    // If you set this to 'foucsed', we will only capture keyboard events
+    // for embdedded decks when they are in focus
+    /* keyboardCondition: null, // ToDo: I think this is not needed */
```
DerMolly, line 105: > Optional function that blocks keyboard events when retuning false
>
> If you set this to 'foucsed', we will only capture keyboard events
> for embdedded decks when they are in focus

Again this enables custom JS functions in the frontmatter. Something we don't want, do we?

src/external-types/reveal.js/index.d.ts
```diff
@@ -4,9 +4,199 @@
...
+    navigationMode?: NavigationMode,
+    // Enable the slide overview mode
+    overview?: boolean,
+    // Flags if it should be possible to pause the presentation (blackout)
+    pause?: boolean,
+    // The maximum number of pages a single slide can expand onto when printing
+    // to PDF, unlimited by default
+    /* pdfMaxPagesPerSlide?: number, // ToDo: I think this is not needed */
+    /* pdfPageHeightOffset: -1, // ToDo: I think this is not needed */
+    /* pdfSeparateFragments: true, // ToDo: I think this is not needed */
```
DerMolly, line 147: We don't include PDF printing so these are unnecessary, aren't they?

src/external-types/reveal.js/index.d.ts
```diff
@@ -4,9 +4,199 @@
...
+    // Exposes the reveal.js API through window.postMessage
+    /* postMessage?: boolean, // ToDo: I think this is not needed */
```
DerMolly, line 149: > Exposes the reveal.js API through window.postMessage

Not helpful, as we don't use reveal for a standalone project, but inside HedgeDoc…

src/external-types/reveal.js/index.d.ts
```diff
@@ -4,9 +4,199 @@
...
+    // Dispatches all reveal.js events to the parent window through postMessage
+    /* postMessageEvents?: boolean, // ToDo: I think this is not needed */
```
DerMolly, line 151: > Dispatches all reveal.js events to the parent window through postMessage

Not helpful, as we don't use reveal for a standalone project, but inside HedgeDoc…

src/external-types/reveal.js/index.d.ts
```diff
@@ -4,9 +4,199 @@
...
+    // - false:  All iframes with data-src will be loaded only when visible
+    preloadIframes?: boolean | null,
+    // Opens links in an iframe preview overlay
+    // Add `data-preview-link` and `data-preview-link="false"` to customise each link
+    // individually
+    previewLinks?: boolean,
+    // Display a presentation progress bar
+    progress?: boolean,
+    // Flags if we should monitor the hash and change slides accordingly
+    /* respondToHashChanges?: boolean, // ToDo: I think this is not needed */
```
DerMolly, line 165: > Flags if we should monitor the hash and change slides accordingly

I think this also depends on, if we want `hash` and therefore `fragmentInURL`…

src/external-types/reveal.js/index.d.ts
```diff
@@ -4,9 +4,199 @@
...
+    // - "h.v":   Horizontal . vertical slide number (default)
+    // - "h/v":   Horizontal / vertical slide number
+    // - "c":   Flattened slide number
+    // - "c/t":   Flattened slide number / total slides
+    //
+    // Alternatively, you can provide a function that returns the slide
+    // number for the current slide. The function should take in a slide
+    // object and return an array with one string [slideNumber] or
+    // three strings [n1,delimiter,n2]. See #formatSlideNumber().
+    slideNumber?: boolean | SlideNumber,
```
DerMolly, line 191: > Alternatively, you can provide a function that returns the slide
> number for the current slide. The function should take in a slide
> object and return an array with one string [slideNumber] or
> three strings [n1,delimiter,n2]. See #formatSlideNumber().

This is something we don't want, do we?

src/external-types/reveal.js/index.d.ts
DerMolly: It seems all reveal does is provide a custom stylesheet, to print to PDF. The presentation. Maybe we want to keep this around then?

Additional context: https://revealjs.com/pdf-export/

src/external-types/reveal.js/index.d.ts
```diff
@@ -4,9 +4,199 @@
...
+  export type TransitionSpeed = 'default' | 'slow' | 'fast'
+
   export interface RevealOptions {
-    plugins?: string[]
+    autoAnimate?: boolean,
+    // Time between auto-animate transitions
+    autoAnimateDuration?: number,
+    // Auto-animate transition function
+    autoAnimateEasing?: 'ease', // ToDo: find type for https://developer.mozilla.org/en-US/docs/Web/CSS/easing-function
+    /* autoAnimateMatcher?: null, // ToDo: I think this is not needed */
```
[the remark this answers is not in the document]
ErikMichelson: Yeet it

src/external-types/reveal.js/index.d.ts
ErikMichelson: We should try to implement that on the presentation-mode view.
I used HedgeDoc a few times for presentations and the hotlinking to specific slides was really useful.

src/external-types/reveal.js/index.d.ts
ErikMichelson: That's more or less the same thing as above.

src/external-types/reveal.js/index.d.ts
ErikMichelson: Computer science people start counting with 0, other humans usually start with 1. As HedgeDoc is intended to be also used by non-cs-students, I'd not drop it.

src/external-types/reveal.js/index.d.ts
ErikMichelson: Remove.

src/external-types/reveal.js/index.d.ts
ErikMichelson: We already have this in HedgeDoc 1.x (append `?print-pdf` to a /p/ URL). We should keep it.

src/external-types/reveal.js/index.d.ts
ErikMichelson: Well, it doesn't make sense to have that configurable. But it could be necessary to hardcode that to true for having the editor scroll along the presentation slides.

src/external-types/reveal.js/index.d.ts
ErikMichelson: See above.

src/external-types/reveal.js/index.d.ts
ErikMichelson: That becomes tricky with react-router, I guess.

src/external-types/reveal.js/index.d.ts
DerMolly: This is true by default

mrdrogdrog: we don't need this.

src/external-types/reveal.js/index.d.ts
mrdrogdrog: This won't work. Inside of the iframe it would only react to changes of the iframe URL.

mrdrogdrog:
Everything that is related to custom functions can be removed.

ErikMichelson (approved):

mrdrogdrog (approved):

PR #1233, opened 2021-05-03
Fix task list checkboxes for numbered lists

ErikMichelson:
### Component/Part
Redux method for updating the checkbox state of a line

### Description
This PR fixes the regex that is used to identify the `[ ]` checkbox part when updating the markdown content for a clicked todo-list item. The previous regex could only detect todo-list items that start with a `-` or `*`. But `+` is a valid unordered list bullet point as well and numbered lists should also be supported.

This PR changes the regex to reflect this and adds tests for the todo-list feature.

### Steps

- [x] Added implementation
- [x] Added / updated tests

### Related Issue(s)
Fixes hedgedoc/hedgedoc#1220

### DIFF
--- a/src/redux/note-details/reducers.ts
+++ b/src/redux/note-details/reducers.ts
@@ -87,7 +87,7 @@ export const NoteDetailsReducer: Reducer<NoteDetails, NoteDetailsAction> = (stat
   }
 }
 
-const TASK_REGEX = /(\s*(?:[-*+]|\d+\.) )(\[[ xX]])( .*)/
+const TASK_REGEX = /(\s*(?:[-*+]|\d+[.)]) )(\[[ xX]])( .*)/
 const setCheckboxInMarkdownContent = (markdownContent: string, lineInMarkdown: number, checked: boolean): string => {
   const lines = markdownContent.split('\n')
   const results = TASK_REGEX.exec(lines[lineInMarkdown])
--- a/src/redux/note-details/reducers.ts
+++ b/src/redux/note-details/reducers.ts
@@ -87,7 +87,7 @@ export const NoteDetailsReducer: Reducer<NoteDetails, NoteDetailsAction> = (stat
   }
 }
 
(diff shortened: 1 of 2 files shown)


DerMolly:
According to https://github.github.com/gfm/#list-items we should also support list that look like this:
```
10) [ ] test
```
10) [ ] test

DerMolly (changes requested):
see comment above

ErikMichelson:
I didn't know about that syntax. Is included now.

src/redux/note-details/reducers.ts
```diff
@@ -87,7 +87,7 @@ export const NoteDetailsReducer: Reducer<NoteDetails, NoteDetailsAction> = (stat
   }
 }
 
-const TASK_REGEX = /(\s*[-*] )(\[[ xX]])( .*)/
+const TASK_REGEX = /(\s*(?:[-*+]|\d+[.)]) )(\[[ xX]])( .*)/
```
DerMolly, line 90: Also technically any symbol other than ` ` (space) should yield a checked checkbox.

DerMolly (changes requested):

src/redux/note-details/reducers.ts
DerMolly: See https://github.github.com/gfm/#task-list-items-extension-

ErikMichelson: I don't know whether we want this behaviour. I fear it leds to many false-positive recognized strings.

If we want to change it, we'd need to change our markdown-it plugin as well.

DerMolly: I have no strong feelings with this. I just noticed this while scrolling through the spec…

mrdrogdrog: As long as all the ordering characters work I'm fine with it. 🤷

DerMolly (approved):

mrdrogdrog (approved):

PR #1252, opened 2021-05-14
Eject react-scripts and update to webpack 5

mrdrogdrog:
### Component/Part
Project structure

### Description
This PR
- ejects the projects. "Eject" is a command by create-react-app that removes react-scripts from the dependencies and copies it's own dependencies and all the create-react-script configuration files for webpack, jest, etc. into the project directory.
- upgrades to webpack 5. This is the main reason for the eject. (I don't want to wait any longer for react-scripts to update to webpack 5)
- updates a lot of dependencies (most of them were outdated because react-script doesn't update them very often)
- replaces a lot of js helper scripts from react-scripts with typescript rewrite/convertions that are executed using ts-node.
- Updates highlight.js to 11-beta to use their own fixed types and get rid of our type hacks
- Moves `src/external-types` to `types`
- Pins the node dev version to 16

### Steps

- [x] Added implementation
- [x] Added / updated tests


DerMolly (approved):

ErikMichelson:
I'm actually a bit unsure about this.
Don't get me wrong, I really appreciate the effort you put into converting and rewriting the helper scripts, @mrdrogdrog!

But the eject means the following: We're now for ourselves in charge of this collection of build and config scripts. With about 20 config scripts and files, this change is quite fundamental. I see the benefits of having the more recent dependency versions (like webpack 5 or the possibility to use ESBuild etc.) but I see the downsides too: Tweaking the configuration isn't a thing of changing one file anymore. Debugging a failing build or struggling during the integration/"merge" with the backend is much more guaranteed to be a headache with these files. How do you explain future contributors our build system consisting of a dozen of files? 

I usually do reviews based on both checking the code and testing out the changes. In this case, I have to admit that I'm probably not able to check the code without investing a larger amount of time, that I currently don't have available.

React (better spoken of CRA of course) isn't magic, so to work for many different projects it is written quite universal. We're a specific project and know our needs quite good (at least that's what I would say). We don't need checking for eight different `.env`-filename variants for example.
My proposal would be the following: Let us return to the basics. Writing our own (or cleaning the existent) webpack-config, integrating only the modules and plugins we really need and extend it with a script that's got the minimum functionality that we need on top of that. This PR would be a good starting point for doing that.

mrdrogdrog:
Tbh that was kinda what i tried. I removed lots of functionality and already tweaked a lot of code. I would love to get rid of react-dev-utils which provides some very useful functions that are a bit buggy and poorly written. Or the "React_app" environment prefix.

There is still a lot to do but the current state shows that it's not that difficult to eject.
I'm still thinking that is when a try.
But I'm not sure if a complete new script structure should be part of this pr.

mrdrogdrog:
IMHO we don't need these training wheels called "create react app" anymore. We're capable to optimize the app in a way we need it.

ErikMichelson:
> IMHO we don't need these training wheels called "create react app" anymore. We're capable of configurating it on our own.

Yes, I fully agree with that and that's literally the essence of my last paragraph. But I'm still not convinced of merging this (in this state) into our code base. These build scripts still contain a lot of unnecessary low-level configuration[1], redundant options and I personally don't see the structure of them.

As mentioned already by @davidmehren, rebuilding the configuration that we need from scratch, structured into logical small units (e.g. common, production, tests) would be great.

> But I'm not sure if a complete new script structure should be part of this pr.

As this PR changes the complete build setup anyway, I'd favor for doing it in this PR too.

---

**Footnotes**

[1]: See the `ts-pnp` package that these scripts use for example. Its readme states:
> ts-pnp is a low-level package - you shouldn't have to use it[...] If you just want to write TypeScript and have it Just Work™, take a look at pnp-webpack-plugin instead.

mrdrogdrog:
About ts-pnp , yeah I want to get rid of that too. Don't worry. I'm not done yet 😉

mrdrogdrog:
This PR is on hold. We use some

mrdrogdrog:
This PR is on hold until the release of react-scripts with webpack 5 because I need some of the changes.

PR #1294, opened 2021-06-03
Add prettier for codestyle and re-format everything

ErikMichelson:
### Component/Part
meta, reformats everything

### Description
This PR adds [prettier](https://prettier.io) as a codestyle enforcer and formatter like the backend repo does. 
This should result in the same clean code over all files. To ensure that prettier is used, husky + lint-staged are included in this PR as well, which run prettier as a pre-commit hook. This is recommended by prettier themselves. Another advantage is that prettier works as a dev-dependency in every node-environment while our editorconfig file only works in JetBrains products.

### Steps

- [x] Added implementation

### Related Issue(s)
none


package.json
```diff
@@ -144,9 +148,21 @@
...
+  "lint-staged": {
+    "src/**/*.{ts,tsx,js}": "prettier -w --ignore-unknown"
+  },
+  "prettier": {
+    "parser": "typescript",
+    "singleQuote": true,
+    "jsxSingleQuote": true,
+    "semi": false,
+    "tabWidth": 2,
+    "trailingComma": "none"
```
DerMolly, line 164: This seems to remove many commas we wrote, maybe this option should not be `none`?

mrdrogdrog: If you check our `.editorconfig` then you'll see that we set `ij_typescript_enforce_trailing_comma = remove`. So I guess we always wanted to remove the commas.

package.json
```diff
@@ -168,9 +184,13 @@
     "cypress": "7.4.0",
     "cypress-commands": "1.1.0",
     "cypress-file-upload": "5.0.7",
+    "eslint-config-prettier": "^8.3.0",
     "eslint-plugin-chai-friendly": "0.7.1",
     "eslint-plugin-cypress": "2.11.3",
     "http-server": "0.12.3",
+    "husky": "^6.0.0",
+    "lint-staged": "^11.0.0",
+    "prettier": "^2.3.0",
```
mrdrogdrog, line 193: pin

src/components/application-loader/initializers/i18n.ts
```diff
@@ -19,7 +19,7 @@ export const setUpI18n = async (frontendAssetsUrl: string): Promise<void> => {
       fallbackLng: 'en',
       debug: process.env.NODE_ENV !== 'production',
       backend: {
-        loadPath: `${ frontendAssetsUrl }locales/{{lng}}.json`
+        loadPath: `${frontendAssetsUrl}locales/{{lng}}.json`
```
mrdrogdrog, line 22: This seems to be wrong. IIRC we WANT these spaces

DerMolly:
I'd suggest adding the following to the prettier config:

```
bracketSpacing: true,
jsxBracketSameLine: true,
arrowParens: 'always',
```

package.json
DerMolly: Then I withdraw my initial comment…

src/components/application-loader/initializers/i18n.ts
ErikMichelson: That's right. I still have to find an option for that though.

DerMolly: Apparently there is no option for this and https://github.com/prettier/prettier/issues/95 suggests prettier won't change this.

DerMolly (approved):

mrdrogdrog (approved):

PR #1319, opened 2021-06-12
Improve stability and speed of E2E tests

mrdrogdrog:
### Component/Part
Cypress E2E CI Jobs

### Description
This PR fixes the stability and speed of the E2E job.

### Steps
- [x] Added / updated tests

### DIFF
--- a/src/components/editor-page/editor-pane/autocompletion/code-block.ts
+++ b/src/components/editor-page/editor-pane/autocompletion/code-block.ts
@@ -6,6 +6,8 @@
 
 import { Editor, Hint, Hints, Pos } from 'codemirror'
 import { findWordAtCursor, generateHintListByPrefix, Hinter } from './index'
+import { DEFAULT_DURATION_IN_SECONDS, dispatchUiNotification } from '../../../../redux/ui-notifications/methods'
+import i18n from 'i18next'
 
 type highlightJsImport = typeof import('../../../common/hljs/hljs')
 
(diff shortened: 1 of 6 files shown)


src/components/editor-page/editor-pane/autocompletion/code-block.ts
```diff
@@ -5,45 +5,103 @@
...
-        }
+/**
+ * Fetches the highlight js chunk.
+ * @return the retrieved highlight js api
+ */
+const loadHighlightJs = async (): Promise<highlightJsImport | null> => {
+  try {
+    return await import('../../../common/hljs/hljs')
+  } catch (error) {
+    console.error("can't load highlight js", error)
```
ErikMichelson, line 23: Maybe use the new notification system instead.

mrdrogdrog: Hm. Maybe. But I'm not sure how much information we should provide. Maybe just "System error occured while doing xyz. Check browser console and report the error if it occurs again." ?
Notifications are for end users.

ErikMichelson: Yes, it's just that end-users get notified that there's a problem and that they may seek help in our matrix chat.

Most people don't look into the browser console if something isn't working and therefore just writing to the console is like a silent failure (which is bad UX).

mrdrogdrog: done

ErikMichelson (approved):

DerMolly (approved):

PR #1329, opened 2021-06-18
Fix unique key in ui-notification-toast

mrdrogdrog:
### Component/Part
Notifications

### Description
Fix unique key error in ui notification

### Steps

- [x] Added implementation

### DIFF
--- a/src/components/notifications/ui-notification-toast.tsx
+++ b/src/components/notifications/ui-notification-toast.tsx
@@ -74,13 +74,13 @@ export const UiNotificationToast: React.FC<UiNotificationProps> = ({
 
   const buttonsDom = useMemo(
     () =>
-      buttons?.map((button) => {
+      buttons?.map((button, buttonIndex) => {
         const buttonClick = () => {
           button.onClick()
           dismissThisNotification()
         }
         return (
-          <Button key={button.label} size={'sm'} onClick={buttonClick} variant={'link'}>
+          <Button key={buttonIndex} size={'sm'} onClick={buttonClick} variant={'link'}>
             {button.label}
           </Button>
         )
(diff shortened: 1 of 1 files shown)


ErikMichelson (approved):

ErikMichelson (approved):

src/components/notifications/ui-notification-toast.tsx
```diff
@@ -74,13 +74,13 @@ export const UiNotificationToast: React.FC<UiNotificationProps> = ({
...
     () =>
-      buttons?.map((button) => {
+      buttons?.map((button, buttonIndex) => {
         const buttonClick = () => {
           button.onClick()
           dismissThisNotification()
         }
         return (
-          <Button key={button.label} size={'sm'} onClick={buttonClick} variant={'link'}>
+          <Button key={buttonIndex} size={'sm'} onClick={buttonClick} variant={'link'}>
```
DerMolly, line 83: Maybe a prefix would be usefull here? Something like `notification-${buttonIndex}`

src/components/notifications/ui-notification-toast.tsx
```diff
@@ -89,9 +89,9 @@ export const UiNotificationToast: React.FC<UiNotificationProps> = ({
   )
 
   const contentDom = useMemo(() => {
-    return content.split('\n').map((value) => {
+    return content.split('\n').map((value, lineNumber) => {
       return (
-        <Fragment>
+        <Fragment key={lineNumber}>
```
DerMolly, line 94: Same here. Additionally doesn't this break if there are more than one notification on the screen at the same time?

src/components/notifications/ui-notification-toast.tsx
mrdrogdrog: Why? Keys only need to be unique within their parent.

src/components/notifications/ui-notification-toast.tsx
mrdrogdrog: s.o.

PR #1346, opened 2021-06-21
Reuse cache for e2e

mrdrogdrog:
### Component/Part
CI e2e tests

### Description
This PR changes the cache key for the build directory in e2e tests to use the commit hash. By doing this the CI don't have to rebuild the same commit if the job gets restarted. 

(It also renames some misnamed steps.)

### Steps

- [x] Added / updated tests

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -28,7 +28,7 @@ jobs:
         id: yarn-cache-dir-path
         run: echo "::set-output name=dir::$(yarn cache dir)"
 
-      - name: Cache node_modules
+      - name: Cache yarn cache
         uses: actions/cache@v2
         id: yarn-cache
         with:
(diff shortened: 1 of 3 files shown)


.github/workflows/e2e.yml
```diff
@@ -20,32 +20,38 @@ jobs:
 
       - name: Cache build
         uses: actions/cache@v2.1.6
+        id: build-cache
         with:
           path: build
-          key: build-${{ hashFiles('src/**') }}
+          key: build-${{ github.sha }}
```
ErikMichelson, line 26: What's the advantage of using the commit sha over the source file hash?

mrdrogdrog: If we use only "src" then the CI would treat commits with the same source but e.g. different package versions as cache hit. Therefore we would have to include every file hash that could influence the build outcome. So if we miss to adjust this key because we e.g. add a new file or env variable then we risk corrupting the e2e test job.
This risk is too high for me. Therefore we should just use the commit hash.
The main purpose of this PR is to speed up tests that get restarted without change.

ErikMichelson: I see, thanks for the explaination.

ErikMichelson (approved):

PR #1413, opened 2021-08-02
Move frontmatter extraction from renderer to redux

### LINKED ISSUE #1228: Move frontMatter to editor
render, edtor

Currently to get any frontmatter options, we render the whole document once and then rerender with the correct options. This is unnecessary. 

Extract the frontmatter in the editor before the first rendering and give the necessary information to the renderer.

Live with one unnecessary render cycle.

ErikMichelson:
### Component/Part
renderer, redux, frontmatter

### Description
This PR moves the extraction and parsing of the frontmatter from the renderer to the editor side. Reasoning is that the renderer should only display things but not handle the logic. The advantage of this move is that frontmatter changes can be applied directly in the editor without sending it forth and back to the renderer, thus reducing delay.
The complete process of content updates and frontmatter extraction is done  with this in the redux core.

### Steps

- [x] Added implementation
- [x] Added / updated tests

### Related Issue(s)
Closes #1228


src/components/editor-page/editor-page.tsx
```diff
@@ -80,6 +82,7 @@ export const EditorPage: React.FC = () => {
   useApplyDarkMode()
   useDocumentTitleWithNoteTitle()
   useEditorModeFromUrl()
+  useExtractFrontmatter(setFrontmatterInfo)
```
mrdrogdrog, line 85: tbh I don't know if I'm happy with this hook. This way the markdown content and the frontmatter information are separate and could be async. You change the markdown content and THEN you adjust the frontmatter based on the markdown. Also you need and editor-page-component. What about the read-only-page?
IMHO a better approach would be to calculate the frontmatter stuff in the redux reducer that sets the new markdown content. This way you make sure, that the markdown content and the note details in the redux state MUST be updated at the same time and are always in sync.
You could also store the cutted frontmatter raw text in the redux store so you know when you have to reparse it.

mrdrogdrog (changes requested):

src/components/editor-page/note-frontmatter/extract-frontmatter.ts
```diff
@@ -0,0 +1,43 @@
...
+ * The block is terminated by a line containing the same amount of dashes or dots as the first line.
+ * @param content The multiline string from which the frontmatter should be extracted.
+ * @return { frontmatterPresent } false if no frontmatter block could be found, true if a block was found.
+ *         { rawFrontmatterText } if a block was found, this property contains the extracted text without the fencing.
+ *         { frontmatterLines }   if a block was found, this property contains the number of lines to skip from the
+ *                                given multiline string for retrieving the non-frontmatter content.
+ */
+export const extractFrontmatter = (content: string): FrontmatterExtractionResult => {
+  const lines = content.split('\n')
+  if (lines.length < 2 || lines[0].length < 3 || lines[0][0] !== '-' || !allCharsEqual(lines[0])) {
```
mrdrogdrog, line 22: I'm confused. I can understand the first conditon. But why are you adding a new util function that does the same job as `lines[0] === '---'` ?

mrdrogdrog (dismissed):

src/components/editor-page/note-frontmatter/extract-frontmatter.ts
ErikMichelson: It doesn't do the same job as you've written.
The `allCharsEqual` method - as stated in it's docs - checks that the given string only consists of the same letters. This condition is there to ensure that `---` is a valid line, but `---test` isn't valid.

src/components/editor-page/editor-page.tsx
ErikMichelson: My initial thought for this was that we could extract the frontmatter parsing into an asynchronously running web-worker later on. But as we discussed it a few days ago, this won't happen because the performance impact isn't worth the web-worker management.

src/components/editor-page/note-frontmatter/extract-frontmatter.ts
mrdrogdrog: Is `----------` also valid as start line?

mrdrogdrog: Just out of curiosity: why don't you just a regex like `^-{3,}$`  ?

src/api/notes/dto-methods.ts
```diff
@@ -7,11 +7,25 @@
 import { NoteDto } from './types'
 import { NoteDetails } from '../../redux/note-details/types'
 import { DateTime } from 'luxon'
-import { initialState } from '../../redux/note-details/reducers'
+import { initialState } from '../../redux/note-details/initial-state'
 
+/**
+ * Converts a note DTO from the API to a {@link NoteDetails} object.
```
mrdrogdrog, line 13:
```suggestion
 * Converts a note DTO from the HTTP API to a {@link NoteDetails} object.
```

src/api/notes/dto-methods.ts
```diff
@@ -7,11 +7,25 @@
...
+ * Note that the documentContent will be set but the markdownContent and rawFrontmatterContent are empty.
+ * Therefore, after getting a note from the API, you need to call the {@link setNoteContent} method.
+ * @param note The NoteDTO as defined in the backend.
+ * @return The NoteDetails object corresponding to the DTO.
+ */
 export const noteDtoToNoteDetails = (note: NoteDto): NoteDetails => {
```
mrdrogdrog, line 19: What I'm noticing now: A verb is missing.

mrdrogdrog, line 19: If I see it correctly then this method is only used in the reducer. And only to overwrite the state with the note from the DTO. Therefore this method should be moved to the reducer.

src/redux/note-details/reducer.ts
```diff
@@ -0,0 +1,167 @@
...
+
+/**
+ * Builds a {@link NoteDetails} redux state from a DTO received as an API response.
+ * @param state The previous redux state.
+ * @param dto The first DTO received from the API containing the relevant information about the note.
+ * @return An updated {@link NoteDetails} redux state.
+ */
+const buildStateFromServerDto = (state: NoteDetails, dto: NoteDto): NoteDetails => {
+  const newState = {
+    ...state,
```
mrdrogdrog, line 44: Why are you preserving the old state here? In fact the convert method overwrites every attribute.

src/api/notes/dto-methods.ts
```diff
@@ -7,11 +7,25 @@
 import { NoteDto } from './types'
 import { NoteDetails } from '../../redux/note-details/types'
 import { DateTime } from 'luxon'
-import { initialState } from '../../redux/note-details/reducers'
+import { initialState } from '../../redux/note-details/initial-state'
 
+/**
+ * Converts a note DTO from the API to a {@link NoteDetails} object.
+ * Note that the documentContent will be set but the markdownContent and rawFrontmatterContent are empty.
+ * Therefore, after getting a note from the API, you need to call the {@link setNoteContent} method.
```
mrdrogdrog, line 15: If this method is used only in the reducer then  maybe we should change it's purpose from a generic "convert" to "createNoteDetailsFromNoteDto"?

Also: If the method is used only in the reducer you can use the util methods to fill "markdownCntent" and "frontmatter"

mrdrogdrog (changes requested):

mrdrogdrog (changes requested):

src/components/common/note-frontmatter/note-frontmatter.ts
```diff
@@ -0,0 +1,76 @@
...
+  tags: string[]
+  deprecatedTagsSyntax: boolean
+  robots: string
+  lang: typeof ISO6391[number]
+  dir: NoteTextDirection
+  breaks: boolean
+  GA: string
+  disqus: string
+  type: NoteType
+  // slideOptions: RevealOptions
```
mrdrogdrog, line 26:
```suggestion
```

src/components/common/note-frontmatter/note-frontmatter.ts
```diff
@@ -0,0 +1,76 @@
...
+      : new Map<string, string>()
+  }
+
+  /**
+   * Creates a new frontmatter metadata instance based on a raw yaml string.
+   * @param content The frontmatter content in yaml format.
+   * @throws Error when the content string is invalid yaml.
+   * @return Frontmatter metadata instance containing the parsed properties from the yaml content.
+   */
+  static parseFromString(content: string): NoteFrontmatter {
```
mrdrogdrog, line 72:
```suggestion
  static createFromYaml(rawYaml: string): NoteFrontmatter {
```

mrdrogdrog (changes requested):

mrdrogdrog (approved):
seal of approval

DerMolly (approved):

PR #1458, opened 2021-08-28
Optimize renovate config

mrdrogdrog:
### Component/Part
Renovate

### Description
This PR changes the renovate config:
- change version range strategy from "always pin" to "autodetect pin" by removing the config extension as the auto detection is already covered by `config:base`. (see https://docs.renovatebot.com/configuration-options/#rangestrategy).
- Remove `dependencyDashboard` as it is already covered by `config:base`

### Steps

- [x] Added implementation

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -2,8 +2,7 @@
   "extends": [
     "config:base",
     ":gitSignOff",
-    ":maintainLockFilesWeekly",
-    ":dependencyDashboard"
+    ":maintainLockFilesWeekly"
   ],
   "labels": [
     "type: maintenance"
--- a/renovate.json
+++ b/renovate.json
@@ -3,8 +3,7 @@
     "config:base",
     ":gitSignOff",
     ":maintainLockFilesWeekly",
-    ":dependencyDashboard",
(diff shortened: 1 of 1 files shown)


ErikMichelson:
I'm not sure whether the `auto` setting is the best way to handle dependencies.
Didn't we want to ensure that every dependency gets pinned? The auto strategy seems to only pin devDependencies and dependencies that are not a library.

mrdrogdrog:
If i understand it correctly then it looks if current repository is a lib or an app.

Also: because we haven't set this value before the auto value has already been used before because this config is included in config:base.

mrdrogdrog:
We could also drop the extra definition because it is already covered by config:base. But don't use "pin all dependencies" because this causes PRs like #1455

ErikMichelson:
If there are sane defaults in `config:base` I see no reason for explicitly defining them again.

mrdrogdrog:
I adjusted the content.

ErikMichelson (approved):

DerMolly (approved):

PR #1612, opened 2021-11-01
Lazy load code highlighting languages

mrdrogdrog:
### Component/Part
HighlightJS

### Description
This PR changes the loading of highlight js language by lazy loading them.

### Steps

- [x] Added implementation


ErikMichelson (commented):
Works fine, looks good.

But maybe we could try to remove the webpack chunk that includes all hljs languages in it and therefore being redundant?

mrdrogdrog:
This webpack chunks comes from the auto completion because it also imports all the languages for the list. I'm not quite sure how to solve this here. Except if we have a static list of languages.

DerMolly:
> Except if we have a static list of languages.

Don't we have this anyway?

mrdrogdrog:
Well. true tho

PR #1636, opened 2021-11-18
Refactor profile page

ErikMichelson:
### Component/Part
Profile page

### Description
This PR refactors the profile page to split long components into multiple smaller ones and to use react hooks like `useCallback` where possible.

### Steps

- [x] Added implementation
- [x] Added / updated tests

### Related Issue(s)
none


src/components/common/countdown-button/countdown-button.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * Button that starts a countdown on render and is only clickable after the countdown has finished.
+ */
+export const CountdownButton: React.FC<CountdownButtonProps> = ({ countdownSeconds, children, ...props }) => {
+  const intervalRef = useRef<NodeJS.Timeout>()
+  const [secondsRemaining, setSecondsRemaining] = useState(countdownSeconds)
+
+  const onTimerTick = useCallback(() => {
+    setSecondsRemaining((previous) => previous - 1)
```
DerMolly, line 23: We could deactivate the interval when this has finished so the counter is zero. It should directly work with the return call of the useEffect blow afaik

public/mock-backend/api/private/tokens
```diff
@@ -1,10 +1,15 @@
 [
     {
         "label": "Demo-App",
-        "created": 1601991518
+        "keyId": "demo"
+        "created": "2021-11-20T23:54:13+01:00",
```
mrdrogdrog, line 5: The types expect "createdAt"

cypress/integration/profile.spec.ts
```diff
@@ -15,7 +15,10 @@ describe('profile page', () => {
         body: [
           {
             label: 'cypress-App',
-            created: 1601991518
+            keyId: 'cypress',
+            created: '2021-11-21T01:11:12+01:00',
```
mrdrogdrog, line 19: The types expect "createdAt"

src/components/profile-page/access-tokens/access-token-creation-form.tsx
```diff
@@ -0,0 +1,132 @@
...
+ */
+export const AccessTokenCreationForm: React.FC = () => {
+  const { t } = useTranslation()
+  const [newTokenLabel, setNewTokenLabel] = useState('')
+  const [newTokenExpiry, setNewTokenExpiry] = useState('')
+  const [newTokenWithSecret, setNewTokenWithSecret] = useState<AccessTokenWithSecret>()
+  const [showCreatedModal, setShowCreatedModal] = useState(false)
+
+  const onHideCratedModal = useCallback(() => {
+    setShowCreatedModal(false)
```
DerMolly, line 29: Should this not clear all states regarding the new token, too?

src/api/tokens/types.d.ts
```diff
@@ -6,9 +6,12 @@
 
 export interface AccessToken {
   label: string
-  created: number
+  validUntil: string
+  keyId: string
+  createdAt: string
+  lastUsed: string
```
mrdrogdrog, line 12: The other date related properties had an adjective.

```suggestion
  lastUsedAt: string
```

src/components/application-loader/initializers/i18n/i18n.ts
```diff
@@ -5,24 +5,24 @@
...
+      import(`../../../../../locales/${language}.json`)
+        .then((resources: ResourceKey) => {
+          callback(null, resources)
+        })
+        .catch((error: Error) => {
+          callback(error, null)
+        })
+    })
+  )
     .use(LanguageDetector)
```
mrdrogdrog, line 26: This is solved in main branch. Please rebase

src/components/profile-page/access-tokens/access-token-deletion-modal.tsx
```diff
@@ -0,0 +1,60 @@
...
+}
+
+/**
+ * Modal that asks for confirmation when deleting an access token.
+ */
+export const AccessTokenDeletionModal: React.FC<AccessTokenDeletionModalProps> = ({ show, token, onHide }) => {
+  useTranslation()
+
+  const onConfirmDelete = useCallback(() => {
+    deleteAccessToken(token.keyId)
```
DerMolly, line 28: We could add a spinner to this, while the request is being processed, but maybe this better kept in an issue for later…

src/components/common/countdown-button/countdown-button.tsx
```diff
@@ -0,0 +1,40 @@
...
+import { Button } from 'react-bootstrap'
+
+export interface CountdownButtonProps extends ButtonProps {
+  countdownSeconds: number
+}
+
+/**
+ * Button that starts a countdown on render and is only clickable after the countdown has finished.
+ */
+export const CountdownButton: React.FC<CountdownButtonProps> = ({ countdownSeconds, children, ...props }) => {
```
mrdrogdrog, line 18: React-Use provides a hook for intervals. See https://github.com/streamich/react-use/blob/master/docs/useInterval.md

src/components/profile-page/settings/profile-change-password.tsx
```diff
@@ -5,11 +5,17 @@
...
 
 import type { ChangeEvent, FormEvent } from 'react'
-import React, { useState } from 'react'
+import React, { useCallback, useState } from 'react'
 import { Button, Card, Form } from 'react-bootstrap'
 import { Trans, useTranslation } from 'react-i18next'
 import { changePassword } from '../../../api/me'
+import { showErrorNotification } from '../../../redux/ui-notifications/methods'
 
+const REGEX_VALID_PASSWORD = /^[^\s].{5,}$/
```
DerMolly, line 14: Is 5 a good minium here? Should we enforce a minium anyway? And if we do shouldn't this be something to be configurable?

src/components/profile-page/access-tokens/access-token-created-modal.tsx
```diff
@@ -0,0 +1,51 @@
...
+import React from 'react'
+import { cypressId } from '../../../utils/cypress-attribute'
+import { Button, Modal } from 'react-bootstrap'
+import { Trans } from 'react-i18next'
+import { CopyableField } from '../../common/copyable/copyable-field/copyable-field'
+import type { CommonModalProps } from '../../common/modals/common-modal'
+import { CommonModal } from '../../common/modals/common-modal'
+import type { AccessTokenWithSecret } from '../../../api/tokens/types'
+
+export interface AccessTokenCreatedModalProps extends Pick<CommonModalProps, 'show' | 'onHide'> {
```
mrdrogdrog, line 16: Redundant use of `Pick<CommonModalProps, 'show' | 'onHide'>`. Maybe we should split the propers and or rename them? There seems to be a use case for Common-Modals that control their own content.

src/components/profile-page/settings/profile-change-password.tsx
DerMolly: If it's configurable, we probably won't be able to give a good error message…

src/components/profile-page/settings/profile-display-name.tsx
```diff
@@ -5,19 +5,23 @@
...
-import { Alert, Button, Card, Form } from 'react-bootstrap'
+import React, { useCallback, useEffect, useState } from 'react'
+import { Button, Card, Form } from 'react-bootstrap'
 import { Trans, useTranslation } from 'react-i18next'
 import { updateDisplayName } from '../../../api/me'
 import { fetchAndSetUser } from '../../login-page/auth/utils'
 import { useApplicationState } from '../../../hooks/common/use-application-state'
+import { showErrorNotification } from '../../../redux/ui-notifications/methods'
 
+const REGEX_INVALID_DISPLAY_NAME = /^\s*$/
```
DerMolly, line 16: maybe it would be desireable to use `REGEX_VALID_DISPLAY_NAME` here, too? For consistencies sake?

DerMolly (approved):

src/components/profile-page/access-tokens/access-token-created-modal.tsx
mrdrogdrog, line 16: Redundant use of `Pick<CommonModalProps, 'show' | 'onHide'>`. Maybe we should split the props and/or rename them? There seems to be a use case for Common-Modals that control their own content.

DerMolly (changes requested):
Accidentaly selected approve, meant to select this.

src/components/profile-page/access-tokens/access-token-creation-form.tsx
```diff
@@ -0,0 +1,132 @@
...
+    setNewTokenExpiry(event.target.value)
+  }, [])
+
+  const onCreateToken = useCallback(
+    (event: FormEvent<HTMLFormElement>) => {
+      event.preventDefault()
+      postNewAccessToken(newTokenLabel, newTokenExpiry)
+        .then((tokenWithSecret) => {
+          setNewTokenWithSecret(tokenWithSecret)
+          setShowCreatedModal(true)
```
mrdrogdrog, line 46: Please don't use multiple useStates for states that are tied together! You want to show the token if you have one. Otherwise the modal shouldn't be visible. Every state change triggers a re-render immediately. State are updates are NOT bulked.

src/components/profile-page/access-tokens/access-token-creation-form.tsx
```diff
@@ -0,0 +1,132 @@
...
+      default: today
+        .plus({
+          year: 1
+        })
+        .toISODate()
+    }
+  }, [])
+
+  useEffect(() => {
+    setNewTokenExpiry(expiryDates.default)
```
mrdrogdrog, line 75: Why aren't you using the default function of the useState that you're manipulating here?

mrdrogdrog (changes requested):
First bunch.

src/components/profile-page/access-tokens/access-token-created-modal.tsx
mrdrogdrog: Already done in https://github.com/hedgedoc/react-client/pull/1649

mrdrogdrog: see above

src/components/profile-page/settings/profile-display-name.tsx
```diff
@@ -26,32 +30,30 @@ export const ProfileDisplayName: React.FC = () => {
...
-    doAsyncChange().catch(() => setError(true))
-    event.preventDefault()
-  }
+  const onSubmitNameChange = useCallback(
+    (event: FormEvent) => {
+      event.preventDefault()
+      updateDisplayName(displayName)
+        .then(async () => {
+          await fetchAndSetUser()
+        })
```
mrdrogdrog, line 44: No need for that. You can just return the Promise from fetchAndSetUser

src/components/profile-page/access-tokens/access-token-creation-form.tsx
```diff
@@ -0,0 +1,132 @@
...
+ * Form for creating a new access token.
+ */
+export const AccessTokenCreationForm: React.FC = () => {
+  const { t } = useTranslation()
+  const [newTokenLabel, setNewTokenLabel] = useState('')
+  const [newTokenExpiry, setNewTokenExpiry] = useState('')
+  const [newTokenWithSecret, setNewTokenWithSecret] = useState<AccessTokenWithSecret>()
+  const [showCreatedModal, setShowCreatedModal] = useState(false)
+
+  const onHideCratedModal = useCallback(() => {
```
mrdrogdrog, line 28: Typo

mrdrogdrog (changes requested):

src/components/profile-page/settings/profile-display-name.tsx
ErikMichelson: As this regex does nothing more than ensuring that the display name is not empty, I replaced it with a check `.trim() !== ''`.

src/components/profile-page/settings/profile-display-name.tsx
ErikMichelson: I just copy-and-pasted this snippet from the unrefactored code. I probably should have looked at it.

src/components/common/countdown-button/countdown-button.tsx
ErikMichelson: Fixed by using useInterval as suggested by @mrdrogdrog

src/components/profile-page/access-tokens/access-token-created-modal.tsx
ErikMichelson: Thanks

src/components/profile-page/access-tokens/access-token-creation-form.tsx
ErikMichelson: But how to do that without triggering two additional renderings because of state changes?

src/api/tokens/types.d.ts
ErikMichelson: https://github.com/hedgedoc/hedgedoc/blob/develop/src/auth/auth-token.dto.ts#L19

src/components/profile-page/settings/profile-change-password.tsx
ErikMichelson: I created an issue to discuss this further: hedgedoc/hedgedoc#1846.

src/components/profile-page/access-tokens/access-token-deletion-modal.tsx
ErikMichelson: I created #1651 for this.

src/components/profile-page/access-tokens/access-token-creation-form.tsx
mrdrogdrog: Use only one state

mrdrogdrog: Or using refs (but that makes only sense if you don't want to control the input but read it. Like for password fields)

src/api/tokens/types.d.ts
mrdrogdrog: I'm crying

src/components/common/countdown-button/countdown-button.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ * Button that starts a countdown on render and is only clickable after the countdown has finished.
+ */
+export const CountdownButton: React.FC<CountdownButtonProps> = ({ countdownStartSeconds, children, ...props }) => {
+  const [secondsRemaining, setSecondsRemaining] = useState(countdownStartSeconds)
+
+  useInterval(
+    () => setSecondsRemaining((previous) => previous - 1),
+    useMemo(() => (secondsRemaining <= 0 ? null : 1000), [secondsRemaining])
```
mrdrogdrog, line 24: You can remove the memo here. The value is easy to calculate and the type is a primitive.

src/components/profile-page/access-tokens/access-token-creation-form.tsx
```diff
@@ -0,0 +1,149 @@
...
+  }, [])
+
+  const tokenDataInitialState = useMemo(() => {
+    return {
+      expiryDate: expiryDates.default,
+      label: ''
+    }
+  }, [expiryDates])
+
+  const [newTokenData, setNewTokenData] = useState<NewTokenFormValues>(() => tokenDataInitialState)
```
mrdrogdrog, line 53: :banana: 
imho "data" is a bit generic. How about `formValues`?

src/components/profile-page/access-tokens/access-token-creation-form.tsx
```diff
@@ -0,0 +1,149 @@
...
+        .toISODate()
+    }
+  }, [])
+
+  const tokenDataInitialState = useMemo(() => {
+    return {
+      expiryDate: expiryDates.default,
+      label: ''
+    }
+  }, [expiryDates])
```
mrdrogdrog, line 51: Just curious.. Why does this memo exist? You need it only in the init function of the following state and it will never change. :thinking:

src/components/profile-page/settings/profile-change-password.tsx
```diff
@@ -18,31 +24,42 @@ export const ProfileChangePassword: React.FC = () => {
   const [newPasswordValid, setNewPasswordValid] = useState(false)
   const [newPasswordAgainValid, setNewPasswordAgainValid] = useState(false)
```
mrdrogdrog, line 25: imho it would be nicer to use memos for the valid vars

mrdrogdrog (changes requested):
Your docs are missing parameter annotations

src/components/profile-page/access-tokens/access-token-creation-form.tsx
```diff
@@ -0,0 +1,149 @@
...
+            size='sm'
+            value={newTokenData.expiryDate}
+            className='bg-dark text-light'
+            onChange={onChangeTokenExpiry}
+            min={expiryDates.min}
+            max={expiryDates.max}
+            required
+            {...cypressId('access-token-add-input-expiry')}
+          />
+        </Form.Group>
```
mrdrogdrog, line 132: :banana: 

The form controls (and the following button) are good candidates for component extractions. It would improve the readablity of the whole component because it reduces the loc in this file. You could also extract stuff like onCreateToken into custom hooks.

src/components/profile-page/access-tokens/access-token-creation-form.tsx
```diff
@@ -0,0 +1,149 @@
...
+
+interface NewTokenFormValues {
+  label: string
+  expiryDate: string
+}
+
+/**
+ * Form for creating a new access token.
+ */
+export const AccessTokenCreationForm: React.FC = () => {
```
mrdrogdrog, line 26: :banana: 
This file is very long :/ 140 LOC

The form controls (and the following button) are good candidates for component extractions. Even if the resulting components don't contain any processing logic, it would improve the readablity of this component because it reduces the loc in this file. You could also extract stuff like onCreateToken into custom hooks.

src/components/profile-page/access-tokens/access-token-creation-form.tsx
ErikMichelson: I use it in the `onHideCreatedModal` as well to reset the form value state. The linter suggested using a memo here.

src/components/profile-page/access-tokens/access-token-creation-form.tsx
ErikMichelson: 129 code lines, but yes - I'm changing that.

image

DerMolly (approved):

DerMolly (approved):

mrdrogdrog (approved):

PR #1647, opened 2021-11-20
Fix renderer type test

mrdrogdrog:
### Component/Part
Renderer type test

### Description
This PR fixes the renderer type test

### Steps

- [x] Added implementation
- [x] Added / updated tests

### DIFF
--- a/cypress/support/get-iframe-content.ts
+++ b/cypress/support/get-iframe-content.ts
@@ -4,17 +4,22 @@
  * SPDX-License-Identifier: AGPL-3.0-only
  */
 
+import { RendererType } from '../../src/components/render-page/window-post-message-communicator/rendering-message'
+
 declare namespace Cypress {
   interface Chainable {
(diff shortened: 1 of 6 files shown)


DerMolly (approved):

cypress/support/get-iframe-content.ts
```diff
@@ -4,17 +4,22 @@
  * SPDX-License-Identifier: AGPL-3.0-only
  */
 
+import { RendererType } from '../../src/components/render-page/window-post-message-communicator/rendering-message'
```
ErikMichelson, line 7: Is it good practise to import src-code into tests?

ErikMichelson (approved):

cypress/support/get-iframe-content.ts
mrdrogdrog: Normally: No.
But imho in this case it's okay, because it's a test attribute.

PR #1800, opened 2022-01-25
Add netlify deployments for PRs

### LINKED ISSUE #1799: Run netlify builds for PRs on GH actions
**The problem:**
Until recently, every commit of every PR got build by netlify. Especially on Saturdays with the enormous amount of renovate PRs and rebases, this escalated so that more than 2/3 of our available monthly netlify build-minutes have been used in less than a month.

**The solution:**
We decided to only run builds for the main branch on netlify. This mainly, because netlify has a better buildtime than GitHub actions.
For open (non-draft) pull requests, we'll run a GitHub actions script that does the netlify-build and pushes the artifacts to netlify.
By using the GitHub deployments API, we can show the deployed URL at the PR.

**Alternative:**
Create a netlify plugin that aborts the build when detecting that the build is a draft PR or from renovate. However, this still uses a tiny amount of runner minutes and seems more complex.

ErikMichelson:
### Component/Part
CI/CD netlify

### Description
This PR adds GH actions powered builds for netlify deployments on PRs.

### Steps

- [x] Added implementation

### Related Issue(s)
Closes #1799


.github/workflows/deploy.yml
```diff
@@ -0,0 +1,81 @@
...
+      - name: Get deployment id for usage in the URL
+        id: deploy-url
+        run: echo "::set-output name=url::${{ github.event.number }}--${{ github.head_ref }}"
+
+      - name: Cache yarn cache
+        uses: actions/cache@v2
+        id: yarn-cache
+        with:
+          path: ${{ steps.yarn-cache-dir-path.outputs.dir }}
+          key: ${{ runner.os }}-16-yarn-${{ hashFiles('**/yarn.lock') }}
```
DerMolly, line 32: should this number be hardcoded?

mrdrogdrog: We don't use multiple node versions here. So it can be removed.

mrdrogdrog (changes requested):
Workflow file is invalid

.github/workflows/deploy.yml
ErikMichelson: Well, it's the same as with the lint.yml workflow.

mrdrogdrog (approved):

PR #1957, opened 2022-04-06
Change year in copyright from 2021 to 2022

mrdrogdrog:
### Component/Part
Everything

### Description
This PR changes the year in the copyright of every file to 2022.

### Steps
- [x] Added implementation
- [x] Added / updated tests

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,4 +1,4 @@
-# SPDX-FileCopyrightText: 2021 The HedgeDoc developers (see AUTHORS file)
+# SPDX-FileCopyrightText: 2022 The HedgeDoc developers (see AUTHORS file)
 #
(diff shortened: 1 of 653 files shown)


ErikMichelson (changes requested):
We discussed some time ago to only update the copyright year in files we touched in that year (using the IDE's update copyright on commit function).

mrdrogdrog:
@ErikMichelson 

Okay.. but what about the entries in the reuse dep file?

ErikMichelson:
> Okay.. but what about the entries in the reuse dep file?

I don't have a strong opinion on that. Do it if you want or don't do it if you don't want to.
The problem with changing all files is that seeing the real last change date is quite difficult as everything in the repo has a current timestamp.

PR #1998, opened 2022-04-18
New note creation

### LINKED ISSUE #1234: Create new note with pre-filled-content from url query parameter
https://github.com/hedgedoc/react-client/issues/90#issuecomment-831905553_
->

hedge.doc/new?content=...

When creating a new pad, allow content to be passed in via a query parameter. This would be limited to ~1500 chars because of url limits and encoding bloat, but I assume this would not be that hard to implement ( ala if (queryParams.content && note.content.length == 0) note.content = queryParams.content)

This would be similar to mailto: links which can also contain content.

Use case:

- Provide button on website for students to create a new pad with pre-filled checklist and task to do. (O-Phase: Checkliste "Was sollte ich vor dem Vorlesungsbegin getan haben")

ErikMichelson:
### Component/Part
new note page

### Description
This PR adds the logic/page for creating a new note.
Please note that in the mock this won't work as expected, but with the real backend, the note gets created properly.

Accessing `/new` or clicking the "New"-button both works.
You can also create a note with pre-filled content by accessing a link like this: `/new?content=This%20is%20some%20test%20content`

### Steps

- [x] Added implementation
- [x] Extended changelog

### Related Issue(s)
closes #1234


src/pages/new.tsx
```diff
@@ -0,0 +1,31 @@
...
+ */
+export const NewNotePage: NextPage = () => {
+  const newContent = useSingleStringUrlParameter('content', '')
+
+  const { loading, error, value } = useAsync(() => {
+    return createNote(newContent)
+  }, [newContent])
+
+  return (
+    <AsyncLoadingBoundary loading={loading} componentName={'NewNotePage'} error={error}>
```
mrdrogdrog, line 25: If (for whatever reason) the creation fails then you just get an alert with "An unexpected error occurred while loading 'NewNotePage'.".
What about extending the boundary with a new optional prop that defines an alternative i18n key?

src/pages/api/mock-backend/private/notes/index.ts
```diff
@@ -0,0 +1,59 @@
...
+              canEdit: false
+            }
+          ]
+        }
+      },
+      editedByAtPosition: []
+    },
+    201
+  )
+}
```
mrdrogdrog, line 57: You mock the creation of a note that can't be opened by the frontend (if in mock mode). 
How about returning the meta data of the feature page?

CHANGELOG.md
```diff
@@ -77,6 +77,7 @@ SPDX-License-Identifier: CC-BY-SA-4.0
 - Images that are currently uploading will be rendered as "uploading".
 - Code blocks with `plantuml` as language are rendered as [PlantUML](https://plantuml.com/) diagram using a configured render server.
 - File based motd that supports markdown without html.
+- New notes can be created with a pre-given content when accessing `/new?content=Example%20content`.
```
mrdrogdrog, line 80: I think it would also be useful to let the user define the primary alias via parameter. IIRC the api allows that.

mrdrogdrog (changes requested):

src/pages/new.tsx
ErikMichelson: Changed

src/pages/api/mock-backend/private/notes/index.ts
ErikMichelson: Changed

CHANGELOG.md
ErikMichelson: This does not seem useful as your link including the alias would work only the first time you click it.

DerMolly (approved):

src/components/error-pages/common-error-page.tsx
```diff
@@ -0,0 +1,60 @@
...
+import React, { useMemo } from 'react'
+import type { PropsWithChildren } from 'react'
+import { LandingLayout } from '../landing-layout/landing-layout'
+import { useTranslation } from 'react-i18next'
+
+export interface CommonErrorPageProps {
+  title: string
+  description?: string
+  titleIsI18nKey?: boolean
+  descriptionIsI18nKey?: boolean
```
mrdrogdrog, line 16: Error pages should NEVER be untranslated because they provide user output. Therefore I suggest that these parameters get removed.

mrdrogdrog (changes requested):

src/components/error-pages/common-error-page.tsx
ErikMichelson: Changed

mrdrogdrog (approved):

PR #2026, opened 2022-05-05
Add new loading animation

mrdrogdrog:
### Component/Part
Loading screen

### Description
This PR replaces our loading screen animation with a new custom made.
You can test it by adding a key named "customDelay" with any value to your browser's local storage.

### Steps

- [x] Added implementation


DerMolly:
Not sure I'm overally happy with the gray color…

DerMolly (commented):
Not sure I'm overally happy with the gray color…

DerMolly (approved):

src/components/application-loader/initializers/index.ts
```diff
@@ -12,7 +12,11 @@ import { fetchFrontendConfig } from './fetch-frontend-config'
 import { loadDarkMode } from './load-dark-mode'
 
 const customDelay: () => Promise<void> = async () => {
-  if (window.localStorage.getItem('customDelay')) {
+  if (
+    typeof window !== 'undefined' &&
+    typeof window.localStorage !== 'undefined' &&
+    (window.location.search.startsWith('?customDelay=') || window.localStorage.getItem('customDelay'))
+  ) {
     return new Promise((resolve) => setTimeout(resolve, 5000))
```
ErikMichelson, line 20: Do we want to use the value of the local storage key / the query parameter as the delay to allow other values than 5 seconds?

mrdrogdrog: As you can see: no.

src/components/application-loader/loading-screen/loading-animation.tsx
```diff
@@ -0,0 +1,49 @@
...
+        <IconRow></IconRow>
+        <IconRow></IconRow>
+        <IconRow></IconRow>
+        <IconRow></IconRow>
+        <IconRow></IconRow>
+        <IconRow></IconRow>
+        <IconRow></IconRow>
+        <IconRow></IconRow>
+        <IconRow></IconRow>
+        <IconRow></IconRow>
```
ErikMichelson, line 45: Would it make sense to use `createNumberRangeArray` (and a memo) here?

src/components/application-loader/initializers/index.ts
ErikMichelson: Why not?

mrdrogdrog: It's only a test flag.

ErikMichelson: I want to enjoy my freetime watching the animation and not pressing the reload button every 5 seconds...

ErikMichelson: > It's only a test flag.

Okay then, if you insist on it...

ErikMichelson (approved):

PR #2155, opened 2022-07-01
fix(notification): set position of notification-area to fixed

### LINKED ISSUE #1956: Set fixed position for notifications
**Description**
When performing an action that produces a notification and you're scrolled down, you won't see the notification because notifications are rendered always relative to the document top position.

**To Reproduce**
1. Go to https://hedgedoc.dev/profile
2. Scroll down to "Account management" and try to delete your account (wait the 10 seconds in the modal).
3. The notification is not in the visible viewport (unless you're using a monitor in portrait orientation).
4. Scroll up to the top of the page and see the notification.

**Expected behavior**
See the notification always in the top of the viewport no matter how far you've scrolled.

**Additional context**
position: fixed instead of (inherited) position: relative

DerMolly:
### Component/Part
notifications

### Description
This PR fixes the position of notification-area to fixed.

This prevents notification to stick to the top of the page not able to be seen by users scrolled down the page.

### Steps

- [x] Added implementation

### Related Issue(s)
fixes #1956

### DIFF
--- a/src/components/notifications/notifications.module.scss
+++ b/src/components/notifications/notifications.module.scss
@@ -1,11 +1,11 @@
-/*
- * SPDX-FileCopyrightText: 2021 The HedgeDoc developers (see AUTHORS file)
+/*!
+ * SPDX-FileCopyrightText: 2022 The HedgeDoc developers (see AUTHORS file)
  *
  * SPDX-License-Identifier: AGPL-3.0-only
  */
 
 .notifications-area {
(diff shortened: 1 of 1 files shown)


DerMolly:
I'm currently debating if it would be a good idea to add a e2e test for this or if that would be overkill…

mrdrogdrog (approved):

ErikMichelson (approved):

PR #2173, opened 2022-07-02
Add button to create note if not found

### LINKED ISSUE #1586: Create notes from 404 page
Right now it's not possible to create notes directly using a url.

As mentioned in https://github.com/hedgedoc/hedgedoc/issues/1767 it would be nice to drop the workflow "if you access a free url and the note doesn't exist.. we create it!". So we should add a "Create this note" button on the 404 page.

Reimplement the "if you access a free url and the note doesn't exist.. we create it!" workflow.

mrdrogdrog:
### Component/Part
error message if note is not found

### Description
This PR adds a button on the error page to create a non-existent note

### Steps

- [x] Added implementation

### Related Issue(s)
fixes #1586

### DIFF
--- a/src/api/notes/index.ts
+++ b/src/api/notes/index.ts
@@ -51,7 +51,6 @@ export const createNote = async (markdown: string): Promise<Note> => {
  * @return Content and metadata of the new note.
  */
 export const createNoteWithPrimaryAlias = async (markdown: string, primaryAlias: string): Promise<Note> => {
-  await new Promise((resolve) => setTimeout(resolve, 5000))
   const response = await new PostApiRequestBuilder<Note, void>('notes/' + primaryAlias)
     .withHeader('Content-Type', 'text/markdown')
     .withBody(markdown)
--- a/src/api/notes/index.ts
+++ b/src/api/notes/index.ts
@@ -16,7 +16,7 @@ import { DeleteApiRequestBuilder } from '../common/api-request-builder/delete-ap
  */
 export const getNote = async (noteIdOrAlias: string): Promise<Note> => {
   const response = await new GetApiRequestBuilder<Note>('notes/' + noteIdOrAlias)
-    .withStatusCodeErrorMapping({ 404: 'api.note.notFound', 403: 'api.note.accessDenied' })
+    .withStatusCodeErrorMapping({ 404: 'api.note.notFound', 403: 'api.note.forbidden' })
     .sendRequest()
   return response.asParsedJsonObject()
 }
@@ -51,6 +51,7 @@ export const createNote = async (markdown: string): Promise<Note> => {
  * @return Content and metadata of the new note.
  */
 export const createNoteWithPrimaryAlias = async (markdown: string, primaryAlias: string): Promise<Note> => {
+  await new Promise((resolve) => setTimeout(resolve, 5000))
   const response = await new PostApiRequestBuilder<Note, void>('notes/' + primaryAlias)
     .withHeader('Content-Type', 'text/markdown')
     .withBody(markdown)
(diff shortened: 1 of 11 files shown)


src/api/notes/index.ts
```diff
@@ -51,6 +51,7 @@ export const createNote = async (markdown: string): Promise<Note> => {
  * @return Content and metadata of the new note.
  */
 export const createNoteWithPrimaryAlias = async (markdown: string, primaryAlias: string): Promise<Note> => {
+  await new Promise((resolve) => setTimeout(resolve, 5000))
```
ErikMichelson, line 54: Why did you add this 5 second delay?

src/components/common/note-loading-boundary/create-non-existing-note-hint.tsx
```diff
@@ -0,0 +1,76 @@
...
+import { Alert, Button } from 'react-bootstrap'
+import { useSingleStringUrlParameter } from '../../../hooks/common/use-single-string-url-parameter'
+import { createNoteWithPrimaryAlias } from '../../../api/notes'
+import { useAsyncFn } from 'react-use'
+import { ShowIf } from '../show-if/show-if'
+import { Redirect } from '../redirect'
+import { ForkAwesomeIcon } from '../fork-awesome/fork-awesome-icon'
+import { testId } from '../../../utils/test-id'
+
+export const CreateNonExistingNoteHint: React.FC = () => {
```
ErikMichelson, line 18: ESDoc missing

src/components/common/note-loading-boundary/create-non-existing-note-hint.tsx
```diff
@@ -0,0 +1,76 @@
...
+      <Alert variant={'danger'} {...testId('failedMessage')}>
+        <ForkAwesomeIcon icon={'exclamation-triangle'} className={'mr-1'} />
+        <Trans i18nKey={'noteLoadingBoundary.createNote.error'} />
+      </Alert>
+    )
+  } else {
+    return (
+      <div className={'d-flex flex-column mt-5'}>
+        <b>
+          <Trans i18nKey={'noteLoadingBoundary.createNote.question'} values={{ aliasName: noteIdFromUrl }} />
```
ErikMichelson, line 57: Is the content of `values` cached properly for re-renders, as this is an object and not a plain data type?

ErikMichelson (changes requested):

src/api/notes/index.ts
mrdrogdrog: Testing code

src/components/common/note-loading-boundary/create-non-existing-note-hint.tsx
mrdrogdrog: In this case it should be okay because this part of the component doesn't get rerendered that often.

ErikMichelson: Okay. When we do performance measurements at some point, we could take a look again.

ErikMichelson (commented):
It's only a style question.
But in my opinion the space between the question and the button is too small, considering the large space above the question.

image

Even more, when respecting that the error message doesn't have the large space.

image

mrdrogdrog:
I increased the space between the alert and the headline and also between the button and the text.

The spacing is inteded this way to see which elements are grouped.

mrdrogdrog:
I have another idea. Wait a second

mrdrogdrog:
@ErikMichelson I changed the code to use alerts. This should give enough separation.

ErikMichelson (approved):
Looks much better!

DerMolly (approved):

PR #2174, opened 2022-07-02
Add mutation observer to first heading title extraction logic

### LINKED ISSUE #1816: Race condition between KaTeX and extraction of first header
**Description**
If you edit the first headline, that contains latex code, then the rendered latex symbols won't show up in the document title.
If you edit another line then the title is rendered correctly.

**To Reproduce**
1. Open an empty note.
2. Don't use frontmatter to trigger the "first-heading-is-title" behaviour.
3. Insert a headline that contains latex like `# $\alpha$ alpha`
4. The document title is now "alpha"
5. Add another line and edit it
6. The document title is now "α alpha"
bonus:
7. Edit the headline again.
8. The document title is now "alpha"

**Expected behavior**
The document title should be correct if editing the headline.

**Additional context**
It seems like the KaTeX rendering is async. The extraction logic is triggered directly after the rendering of the note. So if you edit the headline then the extraction logic might be triggered before the async katex actions has finished. If you edit another line then the headline wasn't modified and the symbol is still present and can be extracted.

mrdrogdrog:
### Component/Part
First heading title extraction

### Description
This PR adds an [Mutation Observer](https://developer.mozilla.org/de/docs/Web/API/MutationObserver) to the found first headline element to detect async changes that have been made outside of the react rendering to the DOM-Elements.

### Steps

- [x] Added implementation
- [x] Added / updated tests

### Related Issue(s)
Fixes #1816

### DIFF
--- a/src/components/markdown-renderer/hooks/use-extract-first-headline.ts
+++ b/src/components/markdown-renderer/hooks/use-extract-first-headline.ts
@@ -1,11 +1,12 @@
 /*
- * SPDX-FileCopyrightText: 2021 The HedgeDoc developers (see AUTHORS file)
+ * SPDX-FileCopyrightText: 2022 The HedgeDoc developers (see AUTHORS file)
  *
  * SPDX-License-Identifier: AGPL-3.0-only
  */
 
 import type React from 'react'
-import { useCallback, useRef } from 'react'
+import { useCallback, useEffect, useMemo, useRef } from 'react'
+import { Optional } from '@mrdrogdrog/optional'
 
 /**
  * Extracts the plain text content of a {@link ChildNode node}.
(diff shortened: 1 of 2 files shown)


src/components/markdown-renderer/hooks/use-extract-first-headline.ts
```diff
@@ -53,17 +54,35 @@ export const useExtractFirstHeadline = (
...
+  useEffect(() => () => mutationObserver.disconnect(), [mutationObserver])
+
+  return useCallback(() => {
+    const foundFirstHeading = Optional.ofNullable(documentElement.current)
+      .map((currentDocumentElement) => currentDocumentElement.getElementsByTagName('h1').item(0))
+      .orElse(null)
+    if (foundFirstHeading === currentFirstHeadingElement.current) {
+      return
+    }
+    mutationObserver.disconnect()
```
ErikMichelson, line 81: Why do you terminate the observer here manually? Shouldn't this be done with the disposal of the `useEffect` anyway?

mrdrogdrog: At this point we have a new headline dom-element. Therefore the observer needs to stop observing the old element to avoid side effects.

ErikMichelson: Okay, yes that is understandable.

ErikMichelson (approved):

src/components/markdown-renderer/hooks/use-extract-first-headline.ts
mrdrogdrog: The effects disconnects the mutationobserver if e.g. the components gets unloaded or the callback prop changes.

DerMolly (approved):

Issue #479, opened 2020-08-22
Add spell checker


mrdrogdrog:
**Is your feature request related to a problem? Please describe.**
Codimd 1.6 had a spell checker in the editor, that was able to show misspelled words, but couldn't offer corrections.
The only supported language was english.

**Describe the solution you'd like**
- [ ] Add a fast, code mirror compatible spell checker (maybe https://github.com/cfinke/Typo.js/ ?)
- [ ] Add support for multiple languages
- [ ] Add corrections

**Describe alternatives you've considered**
A banner that says users "buy a dictionary! (link to amazon)"

DerMolly:
Did you investigate how to integrate Typo.js with CodeMirror? All I found was https://github.com/pirtleshell/codemirror-typo#readme and that seems to have no typings (not really a deal breaker as this could easily be added by us).

mrdrogdrog:
I would use https://codemirror.net/demo/lint.html

hydrargyrum:
The [YAML metadata](https://demo.codimd.org/yaml-metadata) supports a `lang` option. It would be good if the spell checker used that option to correct based on the document language.

davidmehren:
I'd love to see support for [LanguageTool](https://languagetool.org/dev), an open-source grammar and style checker. It's available as a SaaS cloud offering and also self-hostable (Java).

lkiesow:
Using LanguageTool as a browser plugin for any input fields works great for most tools and input fields. It would be great if HedgeDoc would just support these browser plugins. That would also allow users to choose whatever service they want to use for this. LaguageTool, Grammarly or whatever they want.

ErikMichelson:
Actually with the usage of CM6 as the editor nowadays, the only thing preventing the usage of LanguageTool or browser native spell checking are the attributes `spellcheck="false"` and `autocorrect="false"` which are set for the CodeMirror container. Removing these attributes seems to work for me without breaking other functionality of the editor.

PR #2236, opened 2022-07-25
feat: add spell check

DerMolly:
### Component/Part
editor

### Description
This PR adds spell check.

### Steps

- [x] Added implementation

### Related Issue(s)
#479

### DIFF
--- a/src/components/editor-page/editor-pane/editor-pane.tsx
+++ b/src/components/editor-page/editor-pane/editor-pane.tsx
@@ -84,6 +84,7 @@ export const EditorPane: React.FC<ScrollProps> = ({ scrollState, onScroll, onMak
   const yjsExtension = useCodeMirrorYjsExtension(yText, awareness)
   const [firstEditorUpdateExtension, firstUpdateHappened] = useOnFirstEditorUpdateExtension()
   useInsertNoteContentIntoYTextInMockModeEffect(firstUpdateHappened, websocketConnection)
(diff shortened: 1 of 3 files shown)


mrdrogdrog:
image

Doesn't look like it's working

DerMolly:
grafik
Seems to be there are two places that this needs to be addeed…

mrdrogdrog:
It seems like it's working.. but.. do we really want that? :sweat_smile:  The browser doesn't really know the context of many words. :/

image

DerMolly:
> Maybe we should provide a method to turn on/off this functionality.

I think this would be the cleanest way of doing it for now. The typo.js and CM6 linter route is something we can always add later

mrdrogdrog:
Then add it to the redux so we can turn it on/off as soon as we have a settings input.

ErikMichelson:
> image
> 
> Doesn't look like it's working

For me its working neither (Chrome 103).

> Maybe we should provide a method to turn on/off this functionality.

Yes, that would be great for this.
 
> Or don't use the native OS typo system and build something with the codemirror linter plugin and typo.js

I think we should rather look whether it would be possible to ignore only specific lines from the browser's typo checker in a later iteration of this feature. Using native methods has the advantage of supporting external tools like LanguageTool as well as not being limited to a list of dictionaries we ship with HedgeDoc.

ErikMichelson (approved):

mrdrogdrog (approved):

PR #2241, opened 2022-07-27
Fixes warnings in some tests

mrdrogdrog:
### Component/Part
Component tests

### Description
This PR fixes some warnings in
- `copy-to-clipboard-button.test.tsx`
- `create-non-existing-note.test.tsx`
- `user-avatar.test.tsx`

### Steps

- [x] Added implementation
- [x] Added / updated tests

### DIFF
--- a/src/utils/wait-for-other-promises-to-finish.ts
+++ b/src/utils/wait-for-other-promises-to-finish.ts
@@ -6,7 +6,12 @@
 
 /**
  * Waits until all other pending promises are processed.
+ *
+ * NodeJS has a queue for async code that waits for being processed. This method adds a promise to the very end of this queue.
+ * If the promise is resolved then this means that all other promises before it have been processed as well.
+ *
+ * @return A promise which resolves when all other promises have been processed
  */
-export async function waitForOtherPromisesToFinish(): Promise<void> {
-  return await new Promise((resolve) => process.nextTick(resolve))
+export function waitForOtherPromisesToFinish(): Promise<void> {
+  return new Promise((resolve) => process.nextTick(resolve))
 }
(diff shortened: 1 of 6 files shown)


DerMolly (approved):

src/utils/wait-for-other-promises-to-finish.ts
```diff
@@ -0,0 +1,12 @@
+/*
+ * SPDX-FileCopyrightText: 2022 The HedgeDoc developers (see AUTHORS file)
+ *
+ * SPDX-License-Identifier: AGPL-3.0-only
+ */
+
+/**
+ * Waits until all other pending promises are processed.
```
ErikMichelson, line 8: I know the context of this method because of your explanations, but this comment alone doesn't seem to be enough for external contributors to me. Especially since the method itself does not explicitly wait for other Promises - that is more of how the method is used.

mrdrogdrog: > Especially since the method itself does not explicitly wait for other Promises

Actually.. it does. I'm awaiting the fulfilment of the next tick. So yes, this method is waiting. But async.

ErikMichelson: Isn't that more of an implementation detail in the Node engine that it works this way?

mrdrogdrog: So what? 
The function name says that it waits until all other promises are processed. After it's done all other promises are processed. I don't see a problem. It does the thing you expect when running it :shrug:

ErikMichelson: Something like that what you wrote above should be included in the ESDoc comment.
> It's awaiting the fulfilment of the next tick. So yes, this method is waiting. But async.

mrdrogdrog: *sigh* here you go

ErikMichelson (approved):

PR #2252, opened 2022-07-30
Use yarn pnp as module linker

mrdrogdrog:
### Component/Part
Package manager

### Description
This PR:
- disables the node_modules legacy mode in yarn and enables yarn pnp
- adds some missing peer dependencies
- replaces the cypress config file, which was a typescript file, with a javascript file because otherwise cypress can't read it anymore.
- Removes netlify-cli from the package.json and uses npx in the deploy scripts because the cli manipulates the node_modules directory and therefore doesn't work anymore if installed via yarn.

### Steps

- [x] Added implementation


.yarnrc.yml
```diff
@@ -1,5 +1,3 @@
-nodeLinker: node-modules
```
ErikMichelson, line 1: Maybe set `nodeLinker: pnp` explicitly to ensure consistency with changes in default behaviour in newer yarn versions?

.github/workflows/e2e.yml
```diff
@@ -66,7 +66,6 @@ jobs:
     needs: build-frontend
     runs-on: ubuntu-latest
     container:
-      # renovate: datasource=docker depName=cypress/browsers versioning=docker
```
ErikMichelson, line 69: Why did you remove the renovate update section for the cypress docker image?

mrdrogdrog: It doesn't work

ErikMichelson (approved):

DerMolly (approved):

PR #2275, opened 2022-08-08
Note Loading Boundary: Replace redirect with refetch from API

mrdrogdrog:
### Component/Part
Note Loading Boundary

### Description
If the note loading boundary creates a note then it tries to redirect the browser to the new URL to force a reload and load the editor. But since the current URL and the redirection URL are the same nothing happens. This PR modifies the boundary to do a refetch from the API instead.

### Steps

- [x] Added implementation
- [x] Added / updated tests

### DIFF
--- a/src/components/common/note-loading-boundary/note-loading-boundary.tsx
+++ b/src/components/common/note-loading-boundary/note-loading-boundary.tsx
@@ -20,11 +20,11 @@ import { ShowIf } from '../show-if/show-if'
  * @param children The react elements that will be shown when the loading was successful.
  */
 export const NoteLoadingBoundary: React.FC<PropsWithChildren<unknown>> = ({ children }) => {
-  const [{ error, loading }, doFetch] = useLoadNoteFromServer()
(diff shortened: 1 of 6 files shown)


src/components/common/note-loading-boundary/note-loading-boundary.tsx
```diff
@@ -20,15 +20,19 @@ import { ShowIf } from '../show-if/show-if'
  * @param children The react elements that will be shown when the loading was successful.
  */
 export const NoteLoadingBoundary: React.FC<PropsWithChildren<unknown>> = ({ children }) => {
-  const { error, loading } = useLoadNoteFromServer()
+  const [{ error, loading }, doFetch] = useLoadNoteFromServer()
+
+  useEffect(() => {
+    doFetch()
```
DerMolly, line 26: This name is not to descriptive. I suggest `loadNoteFromServer`

DerMolly (changes requested):

DerMolly (approved):

ErikMichelson (approved):

PR #2287, opened 2022-08-14
Several fixes from the alpha deployment

ErikMichelson:
### Component/Part
various parts of the editor and API communication

### Description
This PR fixes several things I detected while playing around with alpha.hedgedoc.dev.
The PR is best reviewed commit by commit. Each commit contains either a meaningful title or a short description.

### Steps

- [x] Added implementation
- [ ] Added / updated tests

### Related Issue(s)
none


src/api/notes/index.ts
```diff
@@ -73,5 +73,9 @@ export const createNoteWithPrimaryAlias = async (markdown: string, primaryAlias:
  * @throws {Error} when the api request wasn't successful.
  */
 export const deleteNote = async (noteIdOrAlias: string): Promise<void> => {
-  await new DeleteApiRequestBuilder('notes/' + noteIdOrAlias).sendRequest()
+  await new DeleteApiRequestBuilder<void, NoteDeletionOptions>('notes/' + noteIdOrAlias)
+    .withJsonBody({
+      keepMedia: false // TODO Ask whether the user wants to keep the media uploaded to the note.
```
DerMolly, line 78: Is there an issue for that?

src/redux/history/methods.ts
```diff
@@ -106,15 +106,15 @@ export const toggleHistoryEntryPinning = async (noteId: string): Promise<void> =
   if (!entryToUpdate) {
     return Promise.reject(`History entry for note '${noteId}' not found`)
   }
-  if (entryToUpdate.pinStatus === undefined) {
-    entryToUpdate.pinStatus = false
+  const updatedEntry = {
+    ...entryToUpdate,
+    pinStatus: !entryToUpdate.pinStatus
```
DerMolly, line 111: This uses the casting of undefined to false, is pinStatus now just not able to be undefined?

ErikMichelson: The backend DTO clearly defines a boolean as type.
https://github.com/hedgedoc/hedgedoc/blob/develop/src/history/history-entry.dto.ts#L49

The HistoryEntry in fact also only allows a boolean as type, therefore the if-condition seems to be an unnecessary relict from the past.
https://github.com/hedgedoc/react-client/blob/alpha-fixes/src/api/history/types.ts#L22

src/api/notes/index.ts
ErikMichelson: #2288

DerMolly (approved):

mrdrogdrog:
tbh I'm not quite happy with the notification changes.

Notifications should be used if something happens out of the current workflow and/or is unrelated to the current content.
Like unexpected errors.
"Hurray, you're registered" and "Couldn't register you because your id is already in use" are very predictable messages/errors and shouldn't use notifications.

Another example: We don't use notifications if the yaml-frontmatter is invalid. We show it directly at the context.

We can discuss this if you want but right now I wouldn't merge this as it doesn't fix anything.

ErikMichelson:
The PR contains more than just the notification changes. We can discuss that particular point later, but the other commits are still relevant.

mrdrogdrog:
I've been referring only to the notification commit

src/api/notes/index.ts
mrdrogdrog: why is that issue not refered here?

mrdrogdrog (changes requested):

src/pages/register.tsx
```diff
@@ -45,6 +46,7 @@ export const RegisterPage: NextPage = () => {
     (event: FormEvent) => {
       doLocalRegister(username, displayName, password)
         .then(() => fetchAndSetUser())
+        .then(() => dispatchUiNotification('login.register.success.title', 'login.register.success.message', {}))
```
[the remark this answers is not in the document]
mrdrogdrog: Another possibility would be that you redirect to another page. Then it would be okay for me to use a notification.

src/components/editor-page/document-bar/revisions/revision-list.tsx
```diff
@@ -38,14 +38,16 @@ export const RevisionList: React.FC<RevisionListProps> = ({ selectedRevisionId,
...
-    return revisions.map((revisionListEntry) => (
-      <RevisionListEntry
-        active={selectedRevisionId === revisionListEntry.id}
-        onSelect={() => onRevisionSelect(revisionListEntry.id)}
-        revision={revisionListEntry}
-        key={revisionListEntry.id}
-      />
-    ))
+    return revisions
+      .sort((a, b) => b.id - a.id)
```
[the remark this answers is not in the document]
ErikMichelson: We have a timestamp, but it is an ISO-formatted string. Sorting these would require to calculate the unix timestamp for each of them for comparison.
As this seems quite resource-intensive without greater benefit, I went with the ids as stated in the commit message.

src/api/notes/index.ts
ErikMichelson: Changed

src/components/editor-page/document-bar/revisions/revision-list.tsx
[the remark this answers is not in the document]
mrdrogdrog: The benefit would be that we're independent from the id. We have no guarantee that the id is always descending. It's only purpose is to be unique. If the backend suddenly decides to use uuids then this is broken. 

It don't see why using `DateTime.fromISO` should be `quite resource-intensive`. It's not like we're sorting 100000000 entries every ms. The sorting happens once if you open the list. IMHO there is no need to safe bits and calc cycles of less than a ms.

mrdrogdrog (approved):

PR #2303, opened 2022-08-20
Refactor handling of environment variables

mrdrogdrog:
### Component/Part
Handling of base url environment variables

### Description
This PR introduces environment variables for the editor and renderer base urls.
This is necessary because otherwise the server side renderer can't access these variables. 
This change also prevents that the values of the environment variables get persisted in the code during build time which would make a deployment much more complicated.

### Steps
- [x] Added implementation
- [x] Added / updated tests


.github/workflows/deploy-pr.yml
```diff
@@ -41,7 +41,7 @@ jobs:
         run: yarn install --immutable
 
       - name: Build netlify variant
```
DerMolly, line 43:
```suggestion
      - name: Build mock variant
```

Dockerfile
```diff
@@ -8,7 +8,10 @@ ENV NEXT_TELEMETRY_DISABLED=1
 
 WORKDIR /app
 COPY . ./
-RUN yarn install --immutable && \
+ENV NODE_ENV=production
+RUN rm -rf public/public && \
+    rm -rf src/pages/api && \
```
DerMolly, line 13: Maybe we should comment here why this is done?

package.json
```diff
@@ -3,28 +3,24 @@
...
   "license": "AGPL-3.0",
   "scripts": {
-    "build": "next build",
-    "build:netlify": "cross-env NEXT_PUBLIC_IGNORE_IFRAME_ORIGIN_CONFIG=true yarn build:mock",
-    "build:mock": "cross-env NEXT_PUBLIC_USE_MOCK_API=true next build",
-    "build:test": "cross-env NEXT_PUBLIC_USE_MOCK_API=true NEXT_PUBLIC_TEST_MODE=true next build",
-    "build:for-real-backend": "cross-env NEXT_PUBLIC_USE_MOCK_API=false NEXT_PUBLIC_BACKEND_BASE_URL=/ next build",
+    "build": "cross-env NODE_ENV=production NEXT_PUBLIC_EDITOR_BASE_URL=https://example.org/ next build",
+    "build:mock": "cross-env NEXT_PUBLIC_USE_MOCK_API=true NEXT_PUBLIC_EDITOR_BASE_URL=https://example.org/ next build",
+    "build:test": "cross-env NODE_ENV=test NEXT_PUBLIC_TEST_MODE=true NEXT_PUBLIC_EDITOR_BASE_URL=https://example.org/ next build",
```
DerMolly, line 8: Isn't this problematic? Nobody runs their instance on example.org!

src/components/render-page/iframe-markdown-renderer.tsx
```diff
@@ -128,7 +128,7 @@ export const IframeMarkdownRenderer: React.FC = () => {
   )
 
   if (!baseConfiguration) {
-    return null
+    return <span>This is the render endpoint. If you see this then something is misconfigured :)</span>
```
DerMolly, line 131: This seems a bit vague, does it not?

src/hooks/common/use-base-url.ts
```diff
@@ -0,0 +1,30 @@
...
+export enum ORIGIN {
+  EDITOR,
+  RENDERER,
+  CURRENT_PAGE
+}
+
+/**
+ * Calculates the base url for relative URLs.
+ */
+export const useBaseUrl = (origin = ORIGIN.CURRENT_PAGE): string => {
```
DerMolly, line 20: This will also be used for the backend, correct? But it only works because it's assumed that backend and frontend are running on the same url, right?

mrdrogdrog:
Dont review this yet! I only created a PR because I needed a netlify deployment

src/components/landing-layout/footer/version-info/version-info-link.tsx
```diff
@@ -18,9 +19,14 @@ export const VersionInfoLink: React.FC = () => {
 
   return (
     <Fragment>
-      <a {...cypressId('show-version-modal')} href={'#'} className={'text-light'} onClick={showModal}>
+      <Button
```
ErikMichelson, line 22: What advantage does this have and why is this in this PR?

src/utils/base-url-from-env-extractor.ts
```diff
@@ -0,0 +1,89 @@
...
+export class BaseUrlFromEnvExtractor {
+  private baseUrls: Optional<BaseUrls> | undefined
+  private logger = new Logger('Base URL Configuration')
+
+  private extractUrlFromEnvVar(envVarName: string, envVarValue: string | undefined): Optional<URL> {
+    return Optional.ofNullable(envVarValue)
+      .filter((value) => {
+        const endsWithSlash = value.endsWith('/')
+        if (!endsWithSlash) {
+          this.logger.error(`${envVarName} must end with an '/'`)
```
ErikMichelson, line 24: Why not simply add a trailing slash at this point?

src/utils/test-modes.js
```diff
@@ -0,0 +1,37 @@
+/*
+ * SPDX-FileCopyrightText: 2022 The HedgeDoc developers (see AUTHORS file)
+ *
+ * SPDX-License-Identifier: AGPL-3.0-only
+ */
+
+/**
+ * Checks if the given string is a positive answer (yes, true or 1).
+ *
+ * @param value The value to check
```
ErikMichelson, line 10: In case of plain Javascript add type information to the docs.

src/utils/test-modes.js
```diff
@@ -0,0 +1,37 @@
...
+ */
+const isPositiveAnswer = (value) => {
+  const lowerValue = value.toLowerCase()
+  return lowerValue === 'yes' || lowerValue === '1' || lowerValue === 'true'
+}
+
+/**
+ * Checks if the current runtime is built in e2e test mode.
```
ErikMichelson, line 18: Is no function and therefore does no checking.

Dockerfile
```diff
@@ -9,9 +9,12 @@ ENV NEXT_TELEMETRY_DISABLED=1
...
+    yarn install --immutable && \
     sed -i "s/CLIENT_VERSION_MISSING/${BUILD_VERSION}/" src/version.json && \
-    yarn build:for-real-backend
+    yarn build:mock
```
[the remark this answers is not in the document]
mrdrogdrog: Yes. Actually: no suffix. Changed it.

src/components/landing-layout/footer/version-info/version-info-link.tsx
mrdrogdrog: Links to "#" are broken with the URL changes. Therefore I needed another element to trigger the version modal.

src/utils/base-url-from-env-extractor.ts
mrdrogdrog: I don't wanna "guess" the URL. The admin should set the correct one.

ErikMichelson: I don't think this has anything to do with guessing, it's simply adding the required slash.

Dockerfile
```diff
@@ -9,9 +9,12 @@ ENV NEXT_TELEMETRY_DISABLED=1
 
 WORKDIR /app
 COPY . ./
-RUN yarn install --immutable && \
+ENV NODE_ENV=production
```
ErikMichelson, line 12: The `ENV` statement can be moved one instruction up in the Dockerfile. This increases the possibilities of the build cache to work as even [`ENV` statements create new layers](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#:~:text=Each%20ENV%20line%20creates%20a,following%2C%20and%20then%20building%20it.).

src/utils/base-url-from-env-extractor.ts
mrdrogdrog: It's not "simple"

`https://example.org/hdiwig` isn't a directory URL. It's a file. I have no idea if the admin wants to put the instance under `https://example.org/` or if he is missing a slash.

So..  no. I won't guess 😅

ErikMichelson: Ok

ErikMichelson (approved):

PR #2305, opened 2022-08-21
feat: configure linkify in markdownit with tlds

### LINKED ISSUE #2304: .rocks domains do not get auto-linkified
The...renderer? auto-linkifier?

If I add plain links to my Markdown document using TLDs like .de or .com, they automatically are converted into links. The .rocks TLD is not auto-converted:

image

All links with domains from the [public suffix list](https://publicsuffix.org/list/public_suffix_list.dat) should be auto-linkified.

Keep the current system that provides a confusing user-experience.

**Additional context**

DerMolly:
### Component/Part
renderer

### Description
This PR fixes the autodetection of linkify.

### Steps

- [x] Added implementation
- [x] Added / updated tests

### Related Issue(s)
closes #2304

### DIFF
--- a/package.json
+++ b/package.json
@@ -120,6 +120,7 @@
     "sanitize-filename": "1.6.3",
     "sass": "1.54.5",
     "sharp": "0.30.7",
+    "tlds": "1.231.0",
     "twemoji-colr-font": "14.0.2",
     "uuid": "8.3.2",
     "vega": "5.22.1",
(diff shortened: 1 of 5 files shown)


DerMolly:
grafik

package.json
```diff
@@ -120,6 +120,7 @@
     "sanitize-filename": "1.6.3",
     "sass": "1.54.5",
     "sharp": "0.30.7",
+    "tlds": "^1.231.0",
```
mrdrogdrog, line 123: Pin dependency

mrdrogdrog (changes requested):

mrdrogdrog (changes requested):
Something here isn't right.

You added the fix to the rendering hook but are testing it in the test file of an extension.

I suggest that you move the fix into the linkify extension. That makes more sense than putting more stuff into the rendering test.

DerMolly:
Good point. I think this is better?

mrdrogdrog (approved):

PR #2322, opened 2022-08-27
fix(websocket): error when leaving realtime-note

ErikMichelson:
### Component/Part
realtime communication

### Description
When a user left a realtime-note, the awareness tried to send events even after the socket was closed, resulting in the app crashing.
This fix adds the check that the socket is still open before sending any messages.

### Steps

- [x] Added implementation

### Related Issue(s)
none

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
     "@hedgedoc/html-to-react": "1.4.5",
     "@hedgedoc/markdown-it-image-size": "1.0.3",
     "@hedgedoc/markdown-it-task-lists": "1.0.5",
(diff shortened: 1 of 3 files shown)


mrdrogdrog:
Saving the websocket here is more of a workaround than an actual solution because the socket is already saved in the class that throws this error. So I would prefer to add a method to the WebsocketTransporter that tells you if the socket is connected and use THIS here.

DerMolly (approved):

ErikMichelson:
Okay, then this will require hedgedoc/realtime#55 first.

mrdrogdrog (approved):

PR #2329, opened 2022-08-30
Add build version to docker images

ErikMichelson:
### Component/Part
Docker images

### Description
The current alpha deployment does not contain any information about the deployed version.
This PR adds a build argument to the Dockerfile to override the version string. For image built by the GitHub actions CI, it will be set to the commit that is used for the build.

### Steps

- [x] Added implementation

### Related Issue(s)
none

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,11 +4,13 @@
 
 # BUILD
 FROM node:18-alpine AS builder
+ARG BUILD_VERSION=CLIENT_VERSION_MISSING
 ENV NEXT_TELEMETRY_DISABLED=1
 
 WORKDIR /app
 COPY . ./
 RUN yarn install --immutable && \
+    sed -i "s/CLIENT_VERSION_MISSING/${BUILD_VERSION}/" src/version.json && \
     yarn build:for-real-backend
 
 # RUNNER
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -4,11 +4,13 @@
 
 # BUILD
 FROM node:18-alpine AS builder
+ARG BUILD_VERSION=dev
 ENV NEXT_TELEMETRY_DISABLED=1
 
 WORKDIR /app
 COPY . ./
 RUN yarn install --immutable && \
+    sed -i "s/0.0/${BUILD_VERSION}/" src/version.json && \
```
mrdrogdrog, line 13: Why not jq?

mrdrogdrog (changes requested):

Dockerfile
ErikMichelson: Because it is not included in the alpine node image

DerMolly (approved):

mrdrogdrog (approved):

mrdrogdrog:
Approved but please check the remaining comment

Dockerfile
ErikMichelson: Hm, yes. Should that have a special meaning in the rendering in HedgeDoc?

mrdrogdrog: Just something unique. Like "Frontend version not set during build" :yum: 
I mean more unique than "0.0"

PR #2443, opened 2022-10-16
feat(splitter): add snapping splitter

mrdrogdrog:
### Component/Part
Splitter

### Description
This PR implements a new splitter that includes an icon, snapping and buttons for changing to full left/right view.
It also removes the editor view mode. 

### Steps

- [x] Added implementation
- [x] Added / updated tests


mrdrogdrog:
The buttons are part of the grabber element. Therefore you can grab them and move the splitter. It's probably a race condition between the grab (mouse down and drag) and the click (mouse down + mouse up).

DerMolly:
Probably, I notice the problems on a trackpad. I'd argue that the arrow button probably should not be part of the dragable element, this way this problem could be circumvented…

ErikMichelson (commented):
✅ LGTM in chrome

src/components/editor-page/editor-page-content.tsx
```diff
@@ -10,12 +10,9 @@ import { useApplyDarkMode } from '../../hooks/common/use-apply-dark-mode'
 import { updateNoteTitleByFirstHeading } from '../../redux/note-details/methods'
 import { MotdModal } from '../common/motd-modal/motd-modal'
 import { AppBar, AppBarMode } from './app-bar/app-bar'
-import { EditorMode } from './app-bar/editor-view-mode'
-import { useViewModeShortcuts } from './hooks/use-view-mode-shortcuts'
```
DerMolly, line 14: I think we want to keep the keyboard shortcuts…

src/components/editor-page/editor-page-content.tsx
```diff
@@ -10,12 +10,9 @@ import { useApplyDarkMode } from '../../hooks/common/use-apply-dark-mode'
...
 import { Sidebar } from './sidebar/sidebar'
 import { Splitter } from './splitter/splitter'
 import type { DualScrollState, ScrollState } from './synced-scroll/scroll-props'
-import { useEditorModeFromUrl } from './hooks/use-editor-mode-from-url'
```
DerMolly, line 18: If I recall correctly this feature is also necessary to send a specific link that sets the correct editor mode.

src/components/editor-page/splitter/split-divider/split-divider.tsx
```diff
@@ -7,24 +7,61 @@
 import React from 'react'
 import styles from './split-divider.module.scss'
 import { testId } from '../../../../utils/test-id'
+import { Button } from 'react-bootstrap'
+import { ForkAwesomeIcon } from '../../../common/fork-awesome/fork-awesome-icon'
+
+export enum Corners {
```
DerMolly, line 13: Corners is a bad term for this imho. I'd suggest using `SpliiterLocation`, `SpliterState` or something similiar…

src/components/editor-page/splitter/split-divider/split-divider.tsx
```diff
@@ -7,24 +7,61 @@
...
 
 /**
  * Renders the divider between the two editor panes.
  * This divider supports both mouse and touch interactions.
  *
- * @param onGrab The callback, that should be called if the splitter is grabbed.
+ * @param onGrab callback that is triggered if the splitter is grabbed.
+ * @param onLeftButtonClick callback that is triggered when the left arrow button is pressed
+ * @param onRightButtonClick callback that is triggered when the right arrow button is pressed
+ * @param corners defines if the corners on the left or right side should be visible
```
DerMolly, line 36: This parameter should be probably be in singular.

src/components/editor-page/splitter/splitter.tsx
```diff
@@ -48,42 +44,43 @@ const extractHorizontalPosition = (moveEvent: MouseEvent | TouchEvent): number |
   }
 }
 
+const snapPercentage = 10
```
DerMolly, line 47: Maybe this should be
```suggestion
const SNAP_PERCENTAGE = 10
```

DerMolly (changes requested):

src/components/editor-page/splitter/split-divider/split-divider.tsx
```diff
@@ -7,24 +7,63 @@
 import React from 'react'
 import styles from './split-divider.module.scss'
 import { testId } from '../../../../utils/test-id'
+import { Button } from 'react-bootstrap'
+import { ForkAwesomeIcon } from '../../../common/fork-awesome/fork-awesome-icon'
+
+export enum DividerShift {
```
DerMolly, line 13: Would it be a good idea to add explanatory documentation to this enum?

src/components/editor-page/document-bar/share/share-modal.tsx
```diff
@@ -24,18 +24,14 @@ import { useBaseUrl } from '../../../../hooks/common/use-base-url'
...
 
   return (
     <CommonModal show={show} onHide={onHide} showCloseButton={true} title={'editor.modal.shareLink.title'}>
       <Modal.Body>
         <Trans i18nKey={'editor.modal.shareLink.editorDescription'} />
-        <CopyableField
-          content={`${baseUrl}n/${noteIdentifier}?${editorMode}`}
-          shareOriginUrl={`${baseUrl}n/${noteIdentifier}?${editorMode}`}
-        />
+        <CopyableField content={`${baseUrl}n/${noteIdentifier}`} shareOriginUrl={`${baseUrl}n/${noteIdentifier}`} />
```
ErikMichelson, line 34: If you re-add the URL query parameters, they should not be removed here, otherwise a user might not know about them.

ErikMichelson (changes requested):

src/components/editor-page/document-bar/share/share-modal.tsx
mrdrogdrog: After a short discussion I removed the URL query parameter feature completly because we couldn't construct a useful case where this is still needed.

DerMolly (approved):

ErikMichelson (approved):

PR #2452, opened 2022-10-21
fix(ci): run workflows in target repository context

mrdrogdrog:
### Component/Part
CI

### Description
Pull requests from forks can't be run by our CI because of missing permissions. (e.g. https://github.com/hedgedoc/react-client/actions/runs/3295208400/jobs/5436019098)
This PR tries to fix this problem by reconfiguring the workflow so it runs on our repository instead of the fork.

To minimize the risk of secrets getting leaked I also added a label "safe to deploy" that must be present before the deployment code gets executed.

See
- https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs
- https://securitylab.github.com/research/github-actions-preventing-pwn-requests/
- https://github.com/actions/first-interaction/issues/10#issuecomment-1278292860
- https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target

### Steps

- [x] Added implementation


ErikMichelson:
While working on the github action workflows, it might be an idea to add another new label like "force deploy" to trigger the deployment workflow also on PRs that are marked as a draft. 🤔

DerMolly (approved):

.github/workflows/deploy-main.yml
```diff
@@ -8,6 +8,10 @@ on:
   push:
     branches: [ main ]
 
+env:
+  NETLIFY_VERSION: 12.0.11
+  NODE_VERSION: 18
```
ErikMichelson, line 13: Is this recognized by renovate?

ErikMichelson (approved):

.github/workflows/deploy-main.yml
mrdrogdrog: not yet. but we could add a regex for that

PR #2503, opened 2022-11-08
Fix visibility of table-size-picker and emoji-picker overlays

mrdrogdrog:
### Component/Part
Table-Size-Picker & Emoji-Picker

### Description
This PR fixes the visibility and click-away-behaviour of both overlays.

### Steps
- [x] Added implementation


DerMolly (approved):

ErikMichelson (approved):
image

Still a bit weird, but I can live with that and I don't think there's an easy way to get around this.

